PR #18, opened 2018-01-23
Madninja/framed stream client

madninja:
* libp2p_framed_stream:server/4 to avoid having to declare a server/4 in your own module
* libp2p_framed_stream:client to construct a framed stream client around a stream. 
* adds close_state to libp2p_connection to be able to check for a pending remote close

### DIFF
--- a/src/libp2p_connection.erl
+++ b/src/libp2p_connection.erl
@@ -6,19 +6,22 @@
          }).
 
 -type connection() :: #connection{}.
+-type close_state() :: open | closed | pending.
 
--export_type([connection/0]).
+-export_type([connection/0, close_state/0]).
 
 -export([new/2, send/2, send/3,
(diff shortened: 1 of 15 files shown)


Vagabond:
I think this looks good, but maybe we should wait to merge it until we've tested integrating the changes?

Vagabond:
I'm not sure the framed_stream server closes when the connection closes, I'm seeing a lot of leaked processes in my testing.

Vagabond:
Perhaps the framed_stream client/server should link or monitor its connection?

Vagabond:
It may also work to call notify_inert() in yamux_stream when the stream is closing.

Vagabond:
Adding some notify_inert() calls to yamux stream seems to fix the leak.

PR #21, opened 2018-01-25
Make libp2p_swarm:listen_addrs get addresses

madninja:
Fixes transport_tcp to fetch all ip addresses when an all "0" (global)
ip4/ip6 address is used.

This also changes libp2p_swarm:start() and start(0) to use
/ip4/0.0.0.0/tcp/0 as the start address.

Adds a few unit tests for this behavior

### DIFF
--- a/src/libp2p_transport_tcp.erl
+++ b/src/libp2p_transport_tcp.erl
@@ -65,7 +65,10 @@ tcp_listen_addrs(Socket) ->
             [multiaddr({Addr, Port}) ||
                 {_, Opts} <- IFAddrs, {addr, Addr} <- Opts, {flags, Flags} <- Opts,
                 size(Addr) == size(IP),
-                not lists:member(loopback, Flags)]
+                not lists:member(loopback, Flags),
+                %% filter out ipv6 link-local addresses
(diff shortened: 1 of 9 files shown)


src/libp2p_transport_tcp.erl
```diff
@@ -55,6 +54,20 @@ dial(MAddr) ->
...
+    {ok, SockAddr={IP, Port}} = inet:sockname(Socket),
+    case lists:all(fun(D) -> D == 0 end, tuple_to_list(IP)) of
+        false ->
+            [multiaddr(SockAddr)];
+        true ->
+            % all 0 address, collect all non loopback interface addresses
+            {ok, IFAddrs} = inet:getifaddrs(),
+            [multiaddr({Addr, Port}) ||
+                {_, Opts} <- IFAddrs, {addr, Addr} <- Opts, {flags, Flags} <- Opts,
+                size(Addr) == size(IP),
```
Vagabond: Won't this filter out ipv6? we shoyld only filter out ipv6 addresses that begin with 16#fe80 (or `65152` decimal).

Vagabond: Oh, I see now, there's no way to bind to both ipv4 and v6 at the same time.

Issue #26, opened 2018-02-02
Investigate TCP hole punching with SO_REUSEADDR


Vagabond:
The go implementation of libp2p uses REUSEADDR to do TCP hole punching as described here:

http://www.bford.info/pub/net/p2pnat/index.html

This allows you to dial and listen on the same port, which means that, potentially, inbound connections to the NAT on the external port will end up back at the listener.

I think the key thing to change would be to make sure the swarm sets `{reuseaddr, true}` on the TCP sockets and that a dial on the swarm sets the `{port, Port}` option to the same port as the listener for the swarm is bound to. This should be a relatively simple change but, combined with STUN, should allow for more NAT traversals than simply relying on nat-upnp.

madninja:
Fixed with #31

Issue #24, opened 2018-01-28
Integrate nat_upnp


Vagabond:
https://github.com/benoitc/nat_upnp

madninja:
Fixed by #31

PR #31, opened 2018-02-06
Add support for various NAT traversal schemes

Vagabond:
This PR adds the following NAT traversal schemes:

* NAT uPNP support #24 
* NAT-PMP support
* TCP hole punching using reuseaddr/reuseport #26 

Additionally, it uses something like STUN to determine and validate the external IP/port is dialable externally. (note that this last piece doesn't work right now because we need to originate the connection test from a new local port or possibly a new IP).


Vagabond:
OK, this is now rebased and uses the `{unique, true}` option in an appropriate place. It still needs more work:

* 2 unit tests fail
* dialyzer warning about `libp2p_identify:observed_addr` taking an opaque term
* distinguish the various flavors of NAT from each other

Beyond those, we need a TURN/relay design and also a way to do 'mutual dialing' where both sides use the STUN derived external IP/port combination to mutually dial each other to punch a hole in restricted cone NAT.

src/libp2p_transport_tcp.erl
```diff
@@ -39,10 +39,11 @@ start_listener(Sup, Addr, TID) ->
             Options = libp2p_config:get_config(?CONFIG_SECTION, OptionDefaults),
             {TransportOpts, ListenOpts0} =
                 lists:partition(fun({Key, _}) ->
-                                        sets:is_element(Key, TransportKeys)
+                                        sets:is_element(Key, TransportKeys);
+                                   (_) -> true
```
madninja: Can this be removed? What other options are you throwing in there?

src/libp2p_transport_tcp.erl
```diff
@@ -67,16 +68,34 @@ new_connection(Socket) ->
     libp2p_connection:new(?MODULE, #tcp_state{socket=Socket, transport=ranch_tcp}).
 
 -spec dial(string(), [libp2p_swarm:connect_opt()], pos_integer()) -> {ok, libp2p_connection:connection()} | {error, term()}.
-dial(MAddr, _DialOptions, Timeout) ->
+dial(MAddr, DialOptions, Timeout) ->
     case tcp_addr(MAddr) of
         {IP, Port, Type} ->
-            case ranch_tcp:connect(IP, Port, [Type], Timeout) of
+            UniquePort = proplists:get_value(unique_port, DialOptions, false),
```
madninja: Why not just use `unique` here instead of introducing `unique_port`?

src/libp2p_transport_tcp.erl
```diff
@@ -67,16 +68,34 @@ new_connection(Socket) ->
...
+            Options = case Type == inet of
```
madninja: I'm not sure this is right? Type can be `inet` or `inet6`

src/libp2p_transport_tcp.erl
```diff
@@ -85,13 +104,18 @@ tcp_listen_addrs(Socket) ->
...
-            [multiaddr({Addr, Port}) ||
+            ListenAddrs = [multiaddr({Addr, Port}) ||
                 {_, Opts} <- IFAddrs, {addr, Addr} <- Opts, {flags, Flags} <- Opts,
                 size(Addr) == size(IP),
                 not lists:member(loopback, Flags),
                 %% filter out ipv6 link-local addresses
                 not (size(Addr) == 8 andalso element(1, Addr) == 16#fe80)
-            ]
+            ],
+            Parent = self(),
```
madninja: I certainly wasn't thinking of kicking these of here? Perhaps move them to where start_listener happens?

src/libp2p_transport_tcp.erl
Vagabond: The `{raw, ...` option is not a 2-tuple.

src/libp2p_transport_tcp.erl
Vagabond: Unique makes sure we don't reuse an existing session, unique_port makes sure we don't use reuseport. They are different operations that will have different behaviour depending on the type of NAT. Specifically restricted NAT vs port restricted NAT.

madninja: ah, so you can have a unique session and yet reuse a port?

src/libp2p_transport_tcp.erl
madninja: Right the raw option is added below this partition not before?

src/libp2p_transport_tcp.erl
Vagabond: Yes, which would result in the same IP/port combo on both sides, but a *new* socket being opened. This is sort of the magic of reuseport.

Vagabond:
This PR is pretty clean now, the dialyzer failure I believe has to do with the specs around the identify shortcut you added @madninja 

I think we should get this landed and then we can work on the peerbook stuff on top of it.

madninja:
I know it may be a lot but perhaps we can get better coverage for the new bits?

Vagabond:
Open to suggestions on how. I'm not sure how to do it in a remotely portable way...

madninja:
NOTE: I'm going to squash and merge this, to get us back on a common baseline but we need to:

1. Don't have tcp/udp logic in swarm server. The work to plug transports should address this
2. Use peerbook instead of swarm_server to update and maintain peer information

PR #72, opened 2018-08-01
HELIUM-404: PROXY

macpie:
This has to wait until https://github.com/helium/erlang-libp2p/pull/69 is merged


src/proxy/libp2p_proxy_server.erl
```diff
@@ -0,0 +1,186 @@
...
+
+-spec splice(inet:socket(), inet:socket()) -> ok.
+splice(Socket1, Socket2) ->
+    Pid = erlang:spawn(fun() ->
+        receive control_given -> ok end,
+        {ok, FD1} = inet:getfd(Socket1),
+        {ok, FD2} = inet:getfd(Socket2),
+        splicer:splice(FD1, FD2)
+    end),
+    _Ref = erlang:monitor(process, Pid),
```
Vagabond: Maybe use spawn_monitor here instead so this is atomic?

PR #73, opened 2018-08-08
Add a very simple NAT map

Vagabond:
Example from sys.config:

```erlang
    {libp2p, [
        {nat_map, #{{172,31,21,144} => {54,69,103,92}}}
    ]}
```


madninja:
Any chance you could add a quick test for it @Vagabond ?

PR #80, opened 2018-08-23
HELIUM-414: Port 0 re-use

macpie:


test/listen_SUITE.erl
```diff
@@ -61,10 +105,88 @@ already_test(Config) ->
...
+%% @end
+%%--------------------------------------------------------------------
+port0_reuse(_Config) ->
+    SwarmOpts = [{libp2p_transport_tcp, [{nat, false}]}],
+
+    {ok, Swarm} = libp2p_swarm:start(listen_port0_reuse, SwarmOpts),
+    ok = libp2p_swarm:listen(Swarm, "/ip4/0.0.0.0/tcp/0"),
+    ok = libp2p_swarm:listen(Swarm, "/ip6/::/tcp/0"),
+
+    timer:sleep(2000),
```
madninja: Why do we need to sleep here?

test/listen_SUITE.erl
```diff
@@ -61,10 +105,88 @@ already_test(Config) ->
...
+    ListenAddrs = libp2p_swarm:listen_addrs(Swarm),
+
+    ok = libp2p_swarm:stop(Swarm),
+    timer:sleep(2000),
```
madninja: Ditto? Is this assuming how long it takes to tear down the swarm? It seems very brittle?

test/listen_SUITE.erl
```diff
@@ -61,10 +105,88 @@ already_test(Config) ->
...
+    {ok, Swarm4} = libp2p_swarm:start(listen_port0_reuse2, SwarmOpts),
+    ok = libp2p_swarm:listen(Swarm4, "/ip4/127.0.0.1/tcp/0"),
+    ok = libp2p_swarm:listen(Swarm4, "/ip6/::1/tcp/0"),
+
+    timer:sleep(2000),
+    ListenAddrs2 = libp2p_swarm:listen_addrs(Swarm4),
+    ok = libp2p_swarm:stop(Swarm4),
+    ok.
+
+%% ------------------------------------------------------------------
```
madninja: Delete the below?

src/libp2p_transport_tcp.erl
```diff
@@ -343,9 +343,14 @@ listen_on(Addr, TID) ->
                                               DefaultListenOpts),
             % Dialyzer severely dislikes ranch_tcp:listen so we
             % emulate it's behavior here
+            Port = reuseport0(TID, Type, IP, Port0),
             case gen_tcp:listen(Port, [Type | AddrOpts] ++ ListenOpts) of
                 {ok, Socket} ->
                     ListenAddrs = tcp_listen_addrs(Socket),
+
+                    Cache = libp2p_swarm_sup:cache(TID),
```
Vagabond: Aren't most of these types of calls wrapped by a function in libp2p_swarm? eg `libp2p_swarm:cache()`

macpie: I did like peer book `libp2p_swarm_sup:register_peerbook` and `libp2p_swarm_sup:peerbook`

madninja: Right but there's _also_ a libp2p_swarm:peerbook function that calls the _sup function

The idea is that libp2p_swarm is the supported API of the swarm. The sup is just an implementation tactic.

Could you add that cache accessor in libp2p_swarm (including an @doc if you could) @macpie ?

PR #82, opened 2018-08-24
Restart stateful transport modules if they exit abnormally

Vagabond:
Note that the installed listeners seem to survive the transport crashing, so this PR mostly focuses on cleanup/restart of the transport itself.

### DIFF
--- a/src/libp2p_transport.erl
+++ b/src/libp2p_transport.erl
@@ -10,9 +10,28 @@
 
 
 -export_type([connection_handler/0]).
--export([for_addr/2, sort_addrs/2, connect_to/4, find_session/3,
+-export([start_link/2, for_addr/2, sort_addrs/2, connect_to/4, find_session/3,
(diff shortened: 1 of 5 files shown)


src/libp2p_transport.erl
```diff
@@ -10,10 +10,29 @@
...
+            libp2p_config:insert_transport(TID, TransportMod, TransportPid),
+            %% on bootup we're blocking the top level supervisor's init, so we need to
+            %% call back asynchronously
+            spawn(fun() ->
+                          Server = libp2p_swarm_sup:server(libp2p_swarm:swarm(TID)),
+                          gen_server:cast(Server, {register, libp2p_config:transport(), TransportPid})
+                  end),
+            {ok, TransportPid};
+        ignore ->
+            %% for some reason we register these transports as `undefined`
```
madninja: We register the transport so `connect` can find the module. The `undefined` is for transports that do not have or need state in a process.

PR #83, opened 2018-08-29
Handle flapping relay peers

Vagabond:
If a relay server is successfully accepting connections, but then
dropping them, a situation we've seen, the relay server will not
increment the relay index to try the next relay peer in the list.

This patch adds a flap counter, so a relay that is flapping will
eventually be bypassed in favor of the next relay in the list, instead
of trying the flapping relay forever.


macpie:
👍

PR #85, opened 2018-08-30
HELIUM-443: NAT

macpie:
This is waiting on `erlang-nat` PR to be merged


src/nat/libp2p_nat.erl
```diff
@@ -0,0 +1,212 @@
...
+%% @doc
+%% @end
+%%--------------------------------------------------------------------
+-spec spawn_discovery(pid(), [string()], ets:tab()) -> ok.
+spawn_discovery(Pid, MultiAddrs, _TID) ->
+    case lists:filtermap(fun discovery_filter/1, MultiAddrs) of
+        [] -> ok;
+        [{MultiAddr, _IP, Port}|_] ->
+            lager:info("trying to map ~p ~p", [MultiAddr, Port]),
+            %% TODO we should make a port mapping for EACH address
```
benoitc: This could be added btw, let me know

Vagabond: Yeah, that would definitely help us as we expect to have our gateway hardware have several network interfaces active simultaneously.

PR #89, opened 2018-09-18
This attempts to offer a closing phase of a relcast_group.

madninja:
The handler for a relcast group can return {close, Timeout}. The
following behavior is epected:

* Stop queing inbound messages
* Stop dialing to maintain outbound connections
* Exit the group after some timeout or after all outbound messages are acked

### DIFF
--- a/src/group/libp2p_group_relcast_server.erl
+++ b/src/group/libp2p_group_relcast_server.erl
@@ -376,7 +376,7 @@ start_workers(TargetAddrs, #state{sup=Sup, group_id=GroupID, tid=TID, self_index
                                           #{ id => make_ref(),
                                              start => {libp2p_group_worker, start_link,
                                                        [Index, ClientSpec, self(), GroupID, TID]},
-                                             restart => permanent
+                                             restart => transient
                                            }),
                       %% sync on the mailbox having been flushed.
                       sys:get_status(WorkerPid),
--- a/src/group/libp2p_group_relcast_server.erl
+++ b/src/group/libp2p_group_relcast_server.erl
@@ -313,8 +313,9 @@ handle_info({send_ack, Index}, State=#state{}) ->
     {noreply, dispatch_ack(Index, State)};
 handle_info(force_close, State=#state{}) ->
     %% The timeout after the handler returned close has fired. Shut
-    %% down the group.
-    {stop, normal, State};
+    %% down the group by exiting the supervisor.
+    exit(State#state.sup, normal),
+    {noreply, State};
 
 handle_info(Msg, State) ->
     lager:warning("Unhandled info: ~p", [Msg]),
--- a/src/group/libp2p_group_relcast_server.erl
+++ b/src/group/libp2p_group_relcast_server.erl
@@ -314,7 +314,7 @@ handle_info({send_ack, Index}, State=#state{}) ->
 handle_info(force_close, State=#state{}) ->
     %% The timeout after the handler returned close has fired. Shut
     %% down the group by exiting the supervisor.
-    exit(State#state.sup, normal),
+    libp2p_swarm:remove_group(State#state.tid, State#state.group_id),
     {noreply, State};
 
 handle_info(Msg, State) ->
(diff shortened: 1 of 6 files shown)


src/group/libp2p_group_relcast_server.erl
```diff
@@ -404,9 +426,33 @@ handle_inbound_message(Index, MsgKey, Msg, State=#state{handler=Handler, handler
...
+    lists:foreach(fun(#worker{pid=self}) ->
+                          ok;
+                     (#worker{pid=Pid}) ->
+                          libp2p_group_worker:close(Pid)
+                  end, Workers),
+    State.
+
+-spec close_check(#state{}) -> #state{}.
+close_check(State=#state{close_state=closing}) ->
+    case count_messages(?OUTBOUND, all, State) of
```
Vagabond: Given we know that no more inbound messages will be added to the queue, can we simply count the remaining messages and decrement a counter, rather than doing an expensive count every time?

Vagabond:
I believe the `libp2p_group_relcast_sup` needs to be changed to allow the children to be 'transient' because I see them being restarted (they're permanent right now).

Presumably whatever the parent of the libp2p_group_relcast_sup is also needs to handle 'normal' exits.

src/group/libp2p_group_relcast_server.erl
madninja: Yeah.. the count is not a traversal over bitcask though, just a fold over the out-keys. You concerned that would take too long?

madninja:
For clarification: The group workers aren't actually stopped individually while the relcast group is closing. They're just in a closing state until they're stopped by the whole relcast group stopping.

@Vagabond are you seeing them being restarted after the relcast_server stops? Or are they being restarted because they crashed for some other reason and I need to handle the closing state as part of group-workers requesting targets?

src/group/libp2p_group_relcast_server.erl
Vagabond: Oh, that's probably fine then.

Vagabond:
Right, but the `libp2p_group_relcast_sup` has 2 permanent children, the server and the worker supervisor. If the server stops normal (when the relcast group is finished), the supervisor restarts the server because its listed as a permanent child, not a transient one.

madninja:
Ah. the relcast_server.. indeed. I thought you meant the workers..

Vagabond:
I think whatever the `libp2p_group_relcast_sup` runs under (which I don't immediately see) should also have it listed as a transient child, so that whole subtree can stop.

madninja:
The group is added as part of a libp2p_swarm:add_group call. It makes the child be temporary.

https://github.com/helium/erlang-libp2p/blob/529877e51a59052b22b00fb194d35d3a392ec43a/src/libp2p_swarm.erl#L327

Vagabond:
Yeah, I just found that using observer, why not transient?

madninja:
Because at the time I wasn't sure that groups actually needed to be restarted when they failed.

Vagabond:
Ok, so now the relcast *server* stops, but the supervisor and the workers stay running (again, according to observer).

Vagabond:
I believe we should call supervisor:terminate_child() on the libp2p_swarm_group_sup, which will kill the relcast_sup for the relcast group. Because it's a temporary child, it will also be removed from the child spec for the supervisor.

Vagabond:
I propose we do something like this:

```diff
diff --git a/src/group/libp2p_group_relcast_server.erl b/src/group/libp2p_group_relcast_server.erl
index c890228..47acf61 100644
--- a/src/group/libp2p_group_relcast_server.erl
+++ b/src/group/libp2p_group_relcast_server.erl
@@ -314,6 +314,7 @@ handle_info({send_ack, Index}, State=#state{}) ->
 handle_info(force_close, State=#state{}) ->
     %% The timeout after the handler returned close has fired. Shut
     %% down the group.
+    supervisor:terminate_child(libp2p_swarm_group_sup:sup(State#state.tid), State#state.group_id),
     {stop, normal, State};

 handle_info(Msg, State) ->
```

PR #92, opened 2018-10-09
Madninja/gossip group

madninja:
This PR tries to address two main issues with gossip groups:

1. Gossip groups only gosiped peerbook entries - This is now fixed by having the new gossip group (there can be only one gossip group per swarm) use a registration approach to have interested parties register for gossip "subjects". 

2. The group worker often misbehaved as it monitored both sessions and streams. The new gossip group has four clear states and only creates and monitors streams associated with it's work. There will be some more work to separate out the actual connectivity parts and property test it. 

3. The gossip group did not account for inbound connections well. The gossip group will now allow up to a maximum number of inbound connections and take those into account when selecting new gossip targets.  Inbound workers are not considered full targets since the gossip group did not pick them. 

4. Sessions are not identified. This goes with the previous point and addresses a potential security hole where a session gossips peerbook entries for the wrong peerbook. Each inbound session is identified to get it's crypto address. Once retrieved the gossip group will allow accepted streams. Since we have this identification we can detect inbound streams and know what their worker target is.


Vagabond:
The process started via `libp2p_peerbook:init()` seems to use a LOT of cpu on startup, followed closely by `libp2p_group_gossip_server:init()`. Are they furiously trying to connect outbound with no backoff delay?

Vagabond:
Testing it with the actual blockchain, I get a lot of crashes like this:

```
2018-10-09 11:16:36 =ERROR REPORT====
** Generic server <0.1391.0> terminating
** Last message in was {inert_read,3,{connection,libp2p_yamux_stream,<0.1390.0>}}
** When Server state == {state,libp2p_gossip_stream,{state,{connection,libp2p_yamux_stream,<0.1390.0>},libp2p_group_gossip_server,<0.1162.0>},server,{connection,libp2p_yamux_stream,<0.1390.0>},#{},<0.1392.0>}
** Reason for termination ==
** {{timeout,{gen_server,call,[<0.1162.0>,{handle_data,"peer",<<10,245,2,10,170,2,10, SNIPPED>>}]}},[{gen_server,call,2,[{file,"gen_server.erl"},{line,215}]},{libp2p_gossip_stream,handle_data,3,[{file,"/home/andrew/erlang-libp2p/_build/test/lib/libp2p/src/group/libp2p_gossip_stream.erl"},{line,58}]},{libp2p_framed_stream,handle_info,2,[{file,"/home/andrew/erlang-libp2p/_build/test/lib/libp2p/src/libp2p_framed_stream.erl"},{line,173}]},{gen_server,try_dispatch,4,[{file,"gen_server.erl"},{line,637}]},{gen_server,handle_msg,6,[{file,"gen_server.erl"},{line,711}]},{libp2p_multistream_server,handle_msg,2,[{file,"/home/andrew/erlang-libp2p/_build/test/lib/libp2p/src/libp2p_multistream_server.erl"},{line,47}]},{proc_lib,init_p_do_apply,3,[{file,"proc_lib.erl"},{line,249}]}]}
```

I think this relates to the high CPU usage above.

Vagabond:
So it looks like group_worker is just clamoring for a peer to connect to:

```
12:40:25.052 [notice] lol worker {in,{cast,{assign_target,{undefined,{"gossip/1.0.0",{libp2p_gossip_stream,[libp2p_group_gossip_server,<0.1228.0>]}}}}}} {<0.1230.0>,targeting}
12:40:25.052 [notice] lol worker {consume,{cast,{assign_target,{undefined,{"gossip/1.0.0",{libp2p_gossip_stream,[libp2p_group_gossip_server,<0.1228.0>]}}}}},targeting} {<0.1230.0>,targeting}
12:40:25.183 [notice] lol worker {in,{info,targeting_timeout}} {<0.1234.0>,targeting}
12:40:25.183 [notice] lol worker {consume,{info,targeting_timeout},targeting} {<0.1234.0>,targeting}
12:40:25.183 [notice] lol peerbook {in,{'$gen_call',{<0.1228.0>,#Ref<0.2997576590.2149842945.241141>},keys}} <0.1229.0>
12:40:25.189 [notice] lol peerbook {out,[<<36,92,4,51,148,147,96,198,86,251,237,123,115,3,6,213,107,120,79,161,240,37,217,205,41,128,159,32,66,233,219,153>>],<0.1228.0>,{state,#Ref<0.2997576590.2149974020.225375>,#Ref<0.2997576590.2149842947.232793>,unknown,300000,#Ref<0.2997576590.2149842948.226002>,86400000,<0.1228.0>,'blockchain_swarm/peerbook',5000,undefined,#{},[],#Fun<libp2p_crypto.0.116422324>}} <0.1229.0>
12:40:25.189 [notice] lol worker {in,{cast,{assign_target,{undefined,{"gossip/1.0.0",{libp2p_gossip_stream,[libp2p_group_gossip_server,<0.1228.0>]}}}}}} {<0.1234.0>,targeting}
12:40:25.189 [notice] lol worker {consume,{cast,{assign_target,{undefined,{"gossip/1.0.0",{libp2p_gossip_stream,[libp2p_group_gossip_server,<0.1228.0>]}}}}},targeting} {<0.1234.0>,targeting}
```

Vagabond:
I think we don't want to send 'undefined' back as a target to the worker. We should let the timer fire and the backoff grow:

```diff
diff --git a/src/group/libp2p_group_gossip_server.erl b/src/group/libp2p_group_gossip_server.erl
index 09374f5..decaa69 100644
--- a/src/group/libp2p_group_gossip_server.erl
+++ b/src/group/libp2p_group_gossip_server.erl
@@ -232,15 +232,16 @@ connections(Kind, #state{workers=Workers}) ->
                 end, [], Workers).

 assign_target(Kind, WorkerPid, TargetAddrs, State=#state{workers=Workers}) ->
-    SelectedAddr = case length(TargetAddrs) of
-                       0 -> undefined;
-                       _ -> mk_multiaddr(lists:nth(rand:uniform(length(TargetAddrs)), TargetAddrs))
-                   end,
-    ClientSpec = {?GROUP_PATH, {libp2p_gossip_stream, [?MODULE, self()]}},
-    libp2p_group_worker:assign_target(WorkerPid, {SelectedAddr, ClientSpec}),
-    NewWorkers = lists:keyreplace(WorkerPid, #worker.pid, Workers,
-                                  #worker{kind=Kind, pid=WorkerPid, target=SelectedAddr}),
-    State#state{workers=NewWorkers}.
+    case length(TargetAddrs) of
+        0 -> State;
+        _ ->
+            SelectedAddr = mk_multiaddr(lists:nth(rand:uniform(length(TargetAddrs)), TargetAddrs)),
+            ClientSpec = {?GROUP_PATH, {libp2p_gossip_stream, [?MODULE, self()]}},
+            libp2p_group_worker:assign_target(WorkerPid, {SelectedAddr, ClientSpec}),
+            NewWorkers = lists:keyreplace(WorkerPid, #worker.pid, Workers,
+                                          #worker{kind=Kind, pid=WorkerPid, target=SelectedAddr}),
+            State#state{workers=NewWorkers}
+    end.

 drop_target(Worker=#worker{pid=WorkerPid}, State=#state{workers=Workers}) ->
     libp2p_group_worker:assign_target(WorkerPid, {undefined, undefined}),
```

Vagabond:
A missing feature seems to be that we don't know WHO gossiped us some data. This would be useful for both knowing who to ask if we need more information about something as well as if the peer is sending us bad/malicious information.

src/group/libp2p_group_gossip_handler.erl
```diff
@@ -0,0 +1,10 @@
+-module(libp2p_group_gossip_handler).
+
+
+-callback init_gossip_data(State::any()) -> init_result().
+-callback handle_gossip_data(Msg::binary(), State::any()) -> handler_result().
+
+-type init_result() :: ok | {send, binary()}.
+-type handler_result() :: ok | {send, binary()} | {error, term()}.
```
Vagabond: If I do a `{send, binary()}` return here, where does it go?

Vagabond:
With the patch above, I can grow the blockchain (if I rewrite the gossip handler to use the new callbacks). However, I'm seeing a very high CPU usage by the peerbook and the gossip server.

It looks like every 5 seconds a bunch of new peerbook entries are being written (coming from the gossip server). Over and over again (on a stable 8 node cluster).

src/peerbook/libp2p_peerbook.erl
```diff
@@ -86,6 +92,19 @@ changed_listener(Pid) ->
...
     gen_server:cast(Pid, {update_nat_type, NatType}).
 
+%%
+%% Gossip Group
+%%
+
+-spec handle_gossip_data(binary(), pid()) -> libp2p_group_gossip_handler:handler_result().
+handle_gossip_data(Data, Pid) ->
+    DecodedList = libp2p_peer:decode_list(Data),
+    libp2p_peerbook:put(Pid, DecodedList).
```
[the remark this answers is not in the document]
madninja: Fixed in 06ded43

src/group/libp2p_group_gossip_handler.erl
madninja: heh, right now, nowhere. The idea I had at the time was that the handler fot eh given gossip data could return some more data to send to the gossip group? Is that useful?

madninja:
> With the patch above, I can grow the blockchain (if I rewrite the gossip handler to use the new callbacks). However, I'm seeing a very high CPU usage by the peerbook and the gossip server.
> 
> It looks like every 5 seconds a bunch of new peerbook entries are being written (coming from the gossip server). Over and over again (on a stable 8 node cluster).

I believe my updated `libp2p_peer:supersedes` function is the cause of this. A session identify will enter a peerbook entry at a given time and shortly after an update for that entry will arrive as part of peerbook gossip. This usually happens within the same second. Our timestamp is second accurate which causes it to fire another update. And of it goes into CPU hell. 

I can either make the timestamp more accurate _or_ be more precise about whether a change has actually occurred in `supersedes`. The latter may be tricky since superseding is a non-obvious question to ask of two peer records?

madninja:
> With the patch above, I can grow the blockchain (if I rewrite the gossip handler to use the new callbacks). However, I'm seeing a very high CPU usage by the peerbook and the gossip server.
> 
> It looks like every 5 seconds a bunch of new peerbook entries are being written (coming from the gossip server). Over and over again (on a stable 8 node cluster).

I believe my updated `libp2p_peer:supersedes` function is the cause of this. A session identify will enter a peerbook entry at a given time and shortly after an update for that entry will arrive as part of peerbook gossip. This usually happens within the same second. Our timestamp is second accurate which causes it to fire another update. And of it goes into CPU hell. 

I can either make the timestamp more accurate _or_ be more precise about whether a change has actually occurred in `supersedes`. The latter may be tricky since superseding is a non-obvious question to ask of two peer records?

> I think we don't want to send 'undefined' back as a target to the worker. We should let the timer fire and the backoff grow:
> 
> ```diff
> diff --git a/src/group/libp2p_group_gossip_server.erl b/src/group/libp2p_group_gossip_server.erl
> index 09374f5..decaa69 100644
> --- a/src/group/libp2p_group_gossip_server.erl
> +++ b/src/group/libp2p_group_gossip_server.erl
> @@ -232,15 +232,16 @@ connections(Kind, #state{workers=Workers}) ->
>                  end, [], Workers).
> 
>  assign_target(Kind, WorkerPid, TargetAddrs, State=#state{workers=Workers}) ->
> -    SelectedAddr = case length(TargetAddrs) of
> -                       0 -> undefined;
> -                       _ -> mk_multiaddr(lists:nth(rand:uniform(length(TargetAddrs)), TargetAddrs))
> -                   end,
> -    ClientSpec = {?GROUP_PATH, {libp2p_gossip_stream, [?MODULE, self()]}},
> -    libp2p_group_worker:assign_target(WorkerPid, {SelectedAddr, ClientSpec}),
> -    NewWorkers = lists:keyreplace(WorkerPid, #worker.pid, Workers,
> -                                  #worker{kind=Kind, pid=WorkerPid, target=SelectedAddr}),
> -    State#state{workers=NewWorkers}.
> +    case length(TargetAddrs) of
> +        0 -> State;
> +        _ ->
> +            SelectedAddr = mk_multiaddr(lists:nth(rand:uniform(length(TargetAddrs)), TargetAddrs)),
> +            ClientSpec = {?GROUP_PATH, {libp2p_gossip_stream, [?MODULE, self()]}},
> +            libp2p_group_worker:assign_target(WorkerPid, {SelectedAddr, ClientSpec}),
> +            NewWorkers = lists:keyreplace(WorkerPid, #worker.pid, Workers,
> +                                          #worker{kind=Kind, pid=WorkerPid, target=SelectedAddr}),
> +            State#state{workers=NewWorkers}
> +    end.
> 
>  drop_target(Worker=#worker{pid=WorkerPid}, State=#state{workers=Workers}) ->
>      libp2p_group_worker:assign_target(WorkerPid, {undefined, undefined}),
> ```

Hmm.. technically that is correct. However, we also handle `{assign_target. {undefined, _}}` in the targeting state of group_worker, which then backs of. It looks like that is not happening correctly?

Vagabond:
WRT supercedes: Why is a new peerbook entry being created so often? If the peerbook entry only differs by the timestamp, why are we generating a new one?

WRT `{send, binary()}` back from the gossip group, I'm not sure gossiping in reply to a gossip event is a great idea. Maybe remove it for now until we have a use case?

Finally about the timeout - as long as it doesn't go insane with events every 5ms, I don't really mind what the fix is. I was a bit confused with the intent of the code so I wasn't sure what the right fix was. It might be nice to have a more descriptive event than setting the target to undefined?

src/group/libp2p_gossip_stream.erl
```diff
@@ -0,0 +1,63 @@
...
+
+init(server, Connection, [HandlerModule, HandlerState]) ->
+    {ok, Session} = libp2p_connection:session(Connection),
+    State = #state{connection=Connection,
+                   handler_module=HandlerModule,
+                   handler_state=HandlerState},
+    case (catch HandlerModule:accept_stream(HandlerState, Session, self())) of
+        ok -> {ok, State};
+        {error, Reason} -> {stop, {error, Reason}, State};
+        {'EXIT', _} -> {stop, normal, State}
```
Vagabond: why the catch here?

src/group/libp2p_group_gossip_server.erl
```diff
@@ -1,105 +1,214 @@
...
+    {reply, {error, not_ready}, State};
+handle_call({accept_stream, Session, StreamPid}, From, State=#state{}) ->
+    libp2p_session:identify(Session, self(), {From, StreamPid}),
+    {noreply, State};
+handle_call({handle_data, Key, Msg}, _From, State=#state{}) ->
+    %% Incoming message from a gossip stream for a given key
+    case maps:find(Key, State#state.handlers) of
+        error -> {reply, ok, State};
+        {ok, {M, S}} -> case (catch M:handle_gossip_data(Msg, S)) of
+                               {'EXIT', Err} -> {reply, {error, Err}, State};
```
Vagabond: I think the catches should have some explanation as to why they're needed.

src/group/libp2p_group_gossip_server.erl
```diff
@@ -1,105 +1,214 @@
...
+    {noreply, State#state{handlers=maps:remove(Key, Handlers)}};
+
+handle_cast({request_target, inbound, WorkerPid}, State=#state{}) ->
+    {noreply, stop_inbound_worker(WorkerPid, State)};
 handle_cast({request_target, Kind=peerbook, WorkerPid}, State=#state{tid=TID}) ->
-    PeerAddrs = [ libp2p_crypto:address_to_p2p(Key)
-                  || Key <- libp2p_peerbook:keys(libp2p_swarm:peerbook(TID)) ],
+    PeerKeys = try
+                   libp2p_peerbook:keys(libp2p_swarm:peerbook(TID))
+               catch
```
Vagabond: Why would this fail?

src/group/libp2p_group_gossip_server.erl
```diff
@@ -1,105 +1,214 @@
...
     {noreply, assign_target(Kind, WorkerPid, TargetAddrs, State)};
-handle_cast({send, Bin}, State=#state{}) ->
+handle_cast({send, Key, Data}, State=#state{}) ->
     {_, Pids} = lists:unzip(connections(all, State)),
-    lists:foreach(fun(Pid) ->
-                          libp2p_group_worker:send(Pid, ignore, Bin)
-                  end, Pids),
-    {noreply, State};
-handle_cast({send_ready, _Ref, _Ready}, State=#state{}) ->
+    case (catch libp2p_gossip_stream:encode(Key, Data)) of
```
Vagabond: If Key and Data are binaries, how can this fail?

src/group/libp2p_group_gossip_handler.erl
```diff
@@ -0,0 +1,10 @@
+-module(libp2p_group_gossip_handler).
+
+
+-callback init_gossip_data(State::any()) -> init_result().
+-callback handle_gossip_data(Msg::binary(), State::any()) -> handler_result().
+
+-type init_result() :: ok | {send, binary()}.
```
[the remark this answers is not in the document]
Vagabond: Ah, I see below it does.

src/group/libp2p_group_gossip_server.erl
Vagabond: I get it now though, the modules are installed into this handler like a gen_event and you don't want one bad gossip handler breaking the whole gossip server.

src/group/libp2p_group_gossip_server.erl
```diff
@@ -109,41 +218,81 @@ handle_info(Msg, State) ->
...
 
 assign_target(Kind, WorkerPid, TargetAddrs, State=#state{workers=Workers}) ->
-    SelectedAddr = case length(TargetAddrs) of
-                       0 -> undefined;
-                       _ -> mk_multiaddr(lists:nth(rand:uniform(length(TargetAddrs)), TargetAddrs))
-                   end,
-    libp2p_group_worker:assign_target(WorkerPid, SelectedAddr),
-    NewWorkers = lists:keyreplace(WorkerPid, 2, Workers, {Kind, WorkerPid, SelectedAddr}),
-    State#state{workers=NewWorkers}.
+    case length(TargetAddrs) of
```
Vagabond: I thought you didn't like this refactor?

src/group/libp2p_group_worker.erl
```diff
@@ -3,253 +3,306 @@
...
+handle_assign_stream(StreamPid, Data=#data{stream_pid=_CurrentStreamPid}) ->
     %% If send_pid known we have an existing stream. Do not replace.
     case rand:uniform(2) of
         1 ->
             %% lager:debug("Loser stream ~p (addr_info ~p) to assigned stream ~p (addr_info ~p)",
             %%             [StreamPid, libp2p_framed_stream:addr_info(StreamPid),
             %%              _CurrentStreamPid, libp2p_framed_stream:addr_info(_CurrentStreamPid)]),
             libp2p_framed_stream:close(StreamPid),
-            keep_state_and_data;
+            false;
```
Vagabond: What does returning false here do?

src/identify/libp2p_identify.erl
```diff
@@ -1,79 +1,72 @@
...
+                           nonce => binary()
+                         }.
+-export_type([identify/0, identify_map/0]).
+-export([from_map/2, encode/1, decode/1, verify/1,
+         address/1, peer/1, observed_maddr/1, observed_addr/1, nonce/1]).
+
+-spec from_map(identify_map(), libp2p_crypto:sig_fun()) -> identify().
+from_map(Map, SigFun) ->
+    G = fun(K) ->
+                case maps:get(K, Map, no_entry) of
```
Vagabond: maps:get/2 will throw an exception for you, you don't need this wrapper fun...

src/peerbook/libp2p_peerbook.erl
```diff
@@ -107,14 +126,20 @@ init([TID, SigFun]) ->
     Group = group_create(SwarmName),
     Opts = libp2p_swarm:opts(TID),
     StaleTime = libp2p_config:get_opt(Opts, [?MODULE, stale_time], ?DEFAULT_STALE_TIME),
+    GossipGroup = case (catch libp2p_swarm:gossip_group(TID)) of
+                      {'EXIT', _} -> undefined;
```
Vagabond: Why?

Vagabond:
Aside from the above comments, this looks good.

src/identify/libp2p_identify.erl
madninja: Right but it will  throw a badkey instead of a badarg. I was trying to indicate that you're passing in a bad map. Is that going too much against the grain?

src/group/libp2p_gossip_stream.erl
madninja: To stop the noisy crash that happened when the gossip_server was going down during shutdown.

src/group/libp2p_group_gossip_server.erl
madninja: During shutdown the gossip_server could try and reach the peerbook. which causes a loud crash

src/group/libp2p_group_gossip_server.erl
madninja: Key is a string and Data is a binary. I guess it could fail if you passed in ones that aren't?

src/group/libp2p_group_gossip_server.erl
madninja: meh.. 50/50, so I went with it.. it reads better than clearing the target

src/group/libp2p_group_worker.erl
madninja: It returns something that isn't the success case? I don't really rely on it, but figured I didn't want to fall over whatever :close returns

src/peerbook/libp2p_peerbook.erl
madninja: Ah, this was a hack to be able to test the peerbook without a gossip group.. There are a number of tests in the peerbook_SUITE that construct a peerbook outside of a swarm, which this makes a bit easier

src/group/libp2p_group_gossip_server.erl
madninja: I'll add a comment to that effect

src/group/libp2p_group_gossip_server.erl
madninja: I'll add a comment to that effect

src/group/libp2p_group_gossip_server.erl
madninja: Added a comment to that effect (safe guarding the gossip server from bad input)

src/identify/libp2p_identify.erl
madninja: It's going against the grain.. simplifying

src/peerbook/libp2p_peerbook.erl
madninja: Added a comment to help clarify

src/identify/libp2p_identify.erl
```diff
@@ -15,10 +15,7 @@
 -spec from_map(identify_map(), libp2p_crypto:sig_fun()) -> identify().
 from_map(Map, SigFun) ->
     G = fun(K) ->
-                case maps:get(K, Map, no_entry) of
-                    no_entry -> error(badarg);
-                    Value -> Value
-                end
+                maps:get(K, Map)
```
Vagabond: Can you just get rid of the fun and use maps:get directly?

Vagabond:
I'm ok with merging this now, did you want to squash this or anything?

src/identify/libp2p_identify.erl
madninja: my fingers are bleeding from all that typing :-)

madninja: done

PR #93, opened 2018-10-13
WIP Initial integration of new relcast library

Vagabond:
Do not merge yet


src/libp2p_multistream.erl
```diff
@@ -23,23 +23,28 @@ write_lines(Connection, Lines) ->
...
                     {Line, <<>>} = decode_line_body(Data, Size),
                     Line
             end
     end.
 
+
 -spec read_lines(libp2p_connection:connection()) -> [string()] | {error, term()}.
 read_lines(Connection) ->
-    case read_varint(Connection) of
+    case read_varint(Connection, 60000) of
```
madninja: Declare a RECV_TIMEOUT and use that macro here to be consistent with other locations

src/libp2p_multistream_client.erl
```diff
@@ -41,11 +45,19 @@ ls(Connection) ->
...
-    case libp2p_multistream:read(Connection) of
-        Id -> libp2p_multistream:write(Connection, Id);
+    case libp2p_multistream:read(Connection, rand:uniform(20000) + 15000) of
+        Id ->
+            ok = libp2p_multistream:write(Connection, Id);
+        {error, timeout} ->
+            ok = libp2p_multistream:write(Connection, Id),
+            case libp2p_multistream:read(Connection) of
+                Id ->
+                    server_switch;
```
madninja: Could you leave some commentary here our forgetful future selves know why the hell we did all this magic?

src/libp2p_multistream_server.erl
```diff
@@ -19,10 +20,20 @@
...
 
+-spec start_link(libp2p_connection:connection(), [{string(), term()}], any()) -> {ok, pid()}.
+start_link(Connection, Handlers, HandlerOpt) ->
+    {ok, proc_lib:spawn_link(?MODULE, init, [{Connection, Handlers, HandlerOpt}])}.
+
 init({Ref, Connection, Handlers, HandlerOpt}) ->
     ok = libp2p_connection:acknowledge(Connection, Ref),
     self() ! handshake,
-    loop(#state{connection=Connection, handlers=Handlers, handler_opt=HandlerOpt}).
+    TimerRef = erlang:send_after(30000, self(), timeout),
```
madninja: Make or reuse a constant of that 30000 which would help clarify what this does? Is this a RECV_TIMEOUT?

src/libp2p_transport.erl
```diff
@@ -138,6 +138,31 @@ start_client_session(TID, Addr, Connection) ->
...
     case libp2p_multistream_client:negotiate_handler(Handlers, Addr, Connection) of
         {error, Error} -> {error, Error};
+        server_switch ->
+            ChildSpec = #{ id => make_ref(),
+                           start => {libp2p_multistream_server, start_link, [Connection, Handlers, TID]},
+                           restart => temporary,
+                           shutdown => 5000,
+                           type => worker },
+            SessionSup = libp2p_swarm_session_sup:sup(TID),
+            case supervisor:start_child(SessionSup, ChildSpec) of
```
madninja: This seems like  code that was copied from elsehwere? Any way to have that more common? Where did it come from?

src/libp2p_multistream_server.erl
Vagabond: It's the time we allow a session protocol to be negotiated before we kill the connection.

src/libp2p_transport.erl
Vagabond: Its how the multicast_server is initialized, it's not easy to make it generic, I don't think?

PR #94, opened 2018-10-22
Add a liveness timer to yamux_session

madninja:
The livenss timer gets kicked every time the session receives data. If
it ever fires it sends a ping to see if the other side is still
there. If that ping succeeds, it re-kicks the timer, but if it fails
the sesison is closed/stopped.

### DIFF
--- a/src/libp2p_yamux_session.erl
+++ b/src/libp2p_yamux_session.erl
@@ -174,7 +174,8 @@ handle_info(timeout_liveness, State=#state{}) ->
     %% No data received.. Send a ping
     {noreply, ping_send(liveness, State)};
 handle_info(liveness_failed, State=#state{}) ->
-    {stop, {error, liveness}, State};
+    lager:notice("Session liveness failure"),
+    {stop, normal, State};
 
 handle_info({stop, Reason}, State=#state{}) ->
     {stop, Reason, State};
@@ -369,7 +370,7 @@ ping_timeout(PingID, State=#state{pings=Pings}) ->
     case maps:take(PingID, Pings) of
         error -> State;
         {{liveness, _, _}, Pings2} ->
-            %% On a livenes ping timeout,the liveness check has
+            %% On a liveness ping timeout,the liveness check has
             %% failed. We send a message back to ourself to handle the
             %% liveness failure.
             self() ! liveness_failed,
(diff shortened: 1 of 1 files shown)


src/libp2p_yamux_session.erl
```diff
@@ -359,6 +388,12 @@ ping_receive(#header{length=PingID}, State=#state{pings=Pings}) ->
     % with the ping time
     case maps:take(PingID, Pings) of
         error -> State;
+        {{liveness, TimerRef, _}, NewPings} ->
+            %% When we receive a ping sent from a liveness check we
+            %% jus tcancel the timer and kick the liveness timer down
+            %% the road again.
+            erlang:cancel_timer(TimerRef),
```
Vagabond: isn't the cancel here redundant given the kick also cancels it?

src/libp2p_yamux_session.erl
```diff
@@ -359,6 +388,12 @@ ping_receive(#header{length=PingID}, State=#state{pings=Pings}) ->
     % with the ping time
     case maps:take(PingID, Pings) of
         error -> State;
+        {{liveness, TimerRef, _}, NewPings} ->
+            %% When we receive a ping sent from a liveness check we
+            %% jus tcancel the timer and kick the liveness timer down
```
Vagabond: Typo 'jus'

src/libp2p_yamux_session.erl
```diff
@@ -164,6 +170,11 @@ handle_info({send_result, Key, Result}, State=#state{sends=Sends}) ->
 
 handle_info({timeout_ping, PingID}, State=#state{}) ->
     {noreply, ping_timeout(PingID, State)};
+handle_info(timeout_liveness, State=#state{}) ->
+    %% No data received.. Send a ping
+    {noreply, ping_send(liveness, State)};
+handle_info(liveness_failed, State=#state{}) ->
+    {stop, {error, liveness}, State};
```
Vagabond: I'd rather you log and exit normal here, so we avoid a crash report?

src/libp2p_yamux_session.erl
```diff
@@ -341,10 +359,21 @@ handle_send_result({call, From}, Result, State=#state{}) ->
...
+ping_send(From, State=#state{next_ping_id=NextPingID}) ->
+    session_send({ping, From, NextPingID}, header_ping(NextPingID), ?TIMEOUT,
+                 State#state{next_ping_id=NextPingID + 1}).
+
 -spec ping_timeout(ping_id(), #state{}) -> #state{}.
 ping_timeout(PingID, State=#state{pings=Pings}) ->
     case maps:take(PingID, Pings) of
         error -> State;
+        {{liveness, _, _}, Pings2} ->
+            %% On a livenes ping timeout,the liveness check has
```
Vagabond: Typo 'livenes'

src/libp2p_yamux_session.erl
Vagabond: oh sorry, this is another timer.

madninja: It is another timer indeed. Good eye

PR #105, opened 2018-12-01
Add associations to peerbook and gossiped peers

madninja:
See the documentation for libp2p_peerbook:add_association and the CT for usage

### DIFF
--- a/src/peerbook/libp2p_peer.erl
+++ b/src/peerbook/libp2p_peer.erl
@@ -232,7 +232,7 @@ is_similar(Target=#libp2p_signed_peer_pb{peer=#libp2p_peer_pb{}},
         andalso nat_type(Target) == nat_type(Other)
         andalso sets:from_list(listen_addrs(Target)) == sets:from_list(listen_addrs(Other))
         andalso sets:from_list(connected_peers(Target)) == sets:from_list(connected_peers(Other))
-        %% We only comparet the {type, assoc_adddress} parts of an
+        %% We only compare the {type, assoc_adddress} parts of an
         %% association as multiple signatures over the same value will
         %% differ
         andalso sets:from_list(association_addrs(Target)) == sets:from_list(association_addrs(Other)).
--- a/src/peerbook/libp2p_peer.erl
+++ b/src/peerbook/libp2p_peer.erl
@@ -21,7 +21,8 @@
 %% associations
 -export([associations/1, association_addrs/1, associations_set/4, associations_get/2, associations_put/4,
          is_association/3, association_address/1, association_signature/1,
-         mk_association/3, association_verify/2]).
+         mk_association/3, association_verify/2,
+         association_encode/1, association_decode/2]).
 %% metadata (unsigned!)
 -export([metadata/1, metadata_set/2, metadata_put/3, metadata_get/3]).
 %% blacklist (unsigned!)
@@ -201,6 +202,19 @@ association_verify(Assoc=#libp2p_association_pb{}, Address) ->
         true -> true;
         false -> error(invalid_association_signature)
     end.
+
+%% @doc Encodes the given association to it's binary form
+-spec association_encode(association()) -> binary().
+association_encode(Msg=#libp2p_association_pb{}) ->
+    libp2p_peer_pb:encode_msg(Msg).
+
+%% @doc Decodes the given binary to an association and verifies it
+%% against the given peer address.
+-spec association_decode(binary(), PeerAddr::libp2p_crypto:address()) -> binary().
+association_decode(Bin, PeerAddr) ->
+    Msg = libp2p_peer_pb:decode_msg(Bin, libp2p_association_pb),
+    true = association_verify(Msg,PeerAddr),
+    Msg.
 
 %% @doc Returns whether a given `Target' is more recent than `Other'
 -spec supersedes(Target::peer(), Other::peer()) -> boolean().
--- a/src/peerbook/libp2p_peer.erl
+++ b/src/peerbook/libp2p_peer.erl
@@ -210,7 +210,7 @@ association_encode(Msg=#libp2p_association_pb{}) ->
 
 %% @doc Decodes the given binary to an association and verifies it
 %% against the given peer address.
--spec association_decode(binary(), PeerAddr::libp2p_crypto:address()) -> binary().
+-spec association_decode(binary(), PeerAddr::libp2p_crypto:address()) -> association().
 association_decode(Bin, PeerAddr) ->
     Msg = libp2p_peer_pb:decode_msg(Bin, libp2p_association_pb),
     true = association_verify(Msg,PeerAddr),
--- a/src/peerbook/libp2p_peer.erl
+++ b/src/peerbook/libp2p_peer.erl
@@ -213,7 +213,7 @@ association_encode(Msg=#libp2p_association_pb{}) ->
 -spec association_decode(binary(), PeerAddr::libp2p_crypto:address()) -> association().
 association_decode(Bin, PeerAddr) ->
     Msg = libp2p_peer_pb:decode_msg(Bin, libp2p_association_pb),
-    true = association_verify(Msg,PeerAddr),
+    association_verify(Msg,PeerAddr),
     Msg.
 
 %% @doc Returns whether a given `Target' is more recent than `Other'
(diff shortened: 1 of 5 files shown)


src/peerbook/libp2p_peer.proto
```diff
@@ -14,6 +14,12 @@ message peer {
     repeated bytes connected = 3;
     nat_type nat_type = 4;
     int64 timestamp = 5;
+    repeated association associations = 6;
+}
+
+message association {
+    bytes address = 1;
```
Vagabond: It would be nice to support more than one type of association. Eg. if we used libolm for the 'encrypted channels', we could store curve25519 keys in here to allow the bootstrap of the encrypted session.

src/peerbook/libp2p_peerbook.erl
```diff
@@ -217,6 +229,17 @@ handle_cast(changed_listener, State=#state{}) ->
     {noreply, update_this_peer(State)};
 handle_cast({update_nat_type, UpdatedNatType}, State=#state{}) ->
     {noreply, update_this_peer(State#state{nat_type=UpdatedNatType})};
+handle_cast({add_association, AssocAddress, AssocSigFun}, State=#state{associations=Associations}) ->
```
Vagabond: Will these persist across node restarts?

src/peerbook/libp2p_peer.erl
```diff
@@ -93,22 +106,136 @@ metadata_get(Peer=#libp2p_signed_peer_pb{}, Key, Default) ->
...
+-spec association_signature(association()) -> binary().
+association_signature(#libp2p_association_pb{signature=Signature}) ->
+    Signature.
+
+%% @doc Returns true if the association can be verified against the
+%% given peer address. This has to be the same peer address that was
+%% used to construct the signature in the assocation and should be the
+%% address of the peer record containing the given association.
+-spec association_verify(association(), PeerAddr::libp2p_crypto:address()) -> true.
+association_verify(Assoc=#libp2p_association_pb{}, Address) ->
```
Vagabond: Is the user expected to check this, or are peerbook entries checked for this on receipt over gossip?

src/peerbook/libp2p_peer.erl
```diff
@@ -93,22 +106,136 @@ metadata_get(Peer=#libp2p_signed_peer_pb{}, Key, Default) ->
...
 %% timestamp of the peers.
 -spec is_similar(Target::peer(), Other::peer()) -> boolean().
 is_similar(Target=#libp2p_signed_peer_pb{peer=#libp2p_peer_pb{}},
            Other=#libp2p_signed_peer_pb{peer=#libp2p_peer_pb{}}) ->
     address(Target) == address(Other)
         andalso nat_type(Target) == nat_type(Other)
         andalso sets:from_list(listen_addrs(Target)) == sets:from_list(listen_addrs(Other))
-        andalso sets:from_list(connected_peers(Target)) == sets:from_list(connected_peers(Other)).
+        andalso sets:from_list(connected_peers(Target)) == sets:from_list(connected_peers(Other))
+        %% We only comparet the {type, assoc_adddress} parts of an
```
Vagabond: `comparet`

src/peerbook/libp2p_peer.erl
```diff
@@ -186,14 +314,35 @@ decode_list(Bin) ->
 -spec decode(binary()) -> peer().
 decode(Bin) ->
     Msg = libp2p_peer_pb:decode_msg(Bin, libp2p_signed_peer_pb),
-    verify(Msg).
+    true = verify(Msg),
+    Msg.
```
Vagabond: This function can badmatch now, is this expected?

Vagabond: Can't we just let verify() or association_verify() error and let it bubble up?

src/peerbook/libp2p_peer.erl
Vagabond: Maybe this doesn't need to be exported?

src/peerbook/libp2p_peer.erl
```diff
@@ -186,14 +314,35 @@ decode_list(Bin) ->
...
         false -> error(invalid_signature)
     end.
+
+%%
+%% Internal
+%%
+
+-spec sign_peer(#libp2p_peer_pb{}, libp2p_crypto:sig_fun()) -> peer().
+sign_peer(Peer, SigFun) ->
+    EncodedPeer = libp2p_peer_pb:encode_msg(Peer),
```
Vagabond: Is this the signature over the whole peer record? Won't this invalidate the signature if some aspect of the peer changes that is not really germane to whether we control this key? Why not just sign the p2p address?

src/peerbook/libp2p_peerbook.erl
```diff
@@ -95,6 +96,24 @@ changed_listener(Pid) ->
...
+%%
+%% Associations are gossiped with the peer record for the swarm.
+%%
+%% Note that the given association will replace an existing
+%% association with the given type and address of the association.
+-spec add_association(pid(), AssocType::atom(), Assoc::libp2p_peer:association()) -> ok.
+add_association(Pid, AssocType, Assoc) ->
+    gen_server:cast(Pid, {add_association, AssocType, Assoc}).
+
+-spec lookup_association(pid(), AssocType::atom(), AssocAddress::libp2p_crypto:address())
```
Vagabond: a @doc for this function would be nice.

src/peerbook/libp2p_peerbook.erl
```diff
@@ -217,6 +244,13 @@ handle_cast(changed_listener, State=#state{}) ->
     {noreply, update_this_peer(State)};
 handle_cast({update_nat_type, UpdatedNatType}, State=#state{}) ->
     {noreply, update_this_peer(State#state{nat_type=UpdatedNatType})};
+handle_cast({add_association, AssocType, Assoc}, State=#state{}) ->
+    %% Fetch our peer record
+    SwarmAddr = libp2p_swarm:address(State#state.tid),
+    {ok, ThisPeer} = unsafe_fetch_peer(SwarmAddr, State),
+    %% Create the new associatin and put it in the peer
```
Vagabond: Typo

src/peerbook/libp2p_peer.erl
madninja: peerbook entries are verified on:

1. peerbook:put which happens during data gossip
2. data decode,so when a peerbook entry is decoded from binary

madninja: The main reason I exported it was to be symmetric with the higher level verify and to be able to unit test just the association creation logic. Does that sound reasonable?

src/peerbook/libp2p_peer.erl
madninja: hmm.. we sign the whole peer record.. we always did.. since an association adds to the overall peer record we need to sign it again? Were we signing peerbook entries wrong the whole time?

src/peerbook/libp2p_peerbook.erl
madninja: uhm.. it's right there.. two lines above line 109

src/peerbook/libp2p_peer.erl
madninja: ah ok.. I can ignore the response of verify since dialyzer can't catch them all ahead of time

src/peerbook/libp2p_peerbook.erl
Vagabond: no... `lookup_association` has no documentation for it.

src/peerbook/libp2p_peer.erl
Vagabond: Sorry, yes, I misread this change as where we were signing using the associated key, disregard.

src/peerbook/libp2p_peer.erl
Vagabond: yeah, that makes sense. just checking a user shouldn't have to call it

src/peerbook/libp2p_peer.erl
Vagabond: Can or can't?

madninja: Can.. fixed in later commit

src/peerbook/libp2p_peerbook.erl
madninja: Fixed

PR #111, opened 2019-01-03
Replace inert with a pure-Erlang version to avoid socket stealing issues

Vagabond:
This is mainly an interim fix to remove inert while we re-work how connection/stream packets events are done.

Note that since we don't support TLS yet, the comment about TLS sockets is irrelevant.

### DIFF
--- a/src/libp2p_transport_tcp.erl
+++ b/src/libp2p_transport_tcp.erl
@@ -166,10 +166,36 @@ acknowledge(#tcp_state{}, Ref) ->
     ranch:accept_ack(Ref).
 
 -spec fdset(tcp_state()) -> ok | {error, term()}.
-fdset(#tcp_state{socket=Socket}) ->
-    case inet:getfd(Socket) of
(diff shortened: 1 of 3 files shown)


src/libp2p_transport_tcp.erl
```diff
@@ -178,9 +204,24 @@ socket(#tcp_state{socket=Socket}) ->
...
+                true ->
+                    Pid ! clear;
+                false ->
+                    ok
+            end,
+            erlang:put(fdset_pid, undefined)
+    end,
+    %% consume any messages of this form in the mailbox
+    receive
+        {inert_read, 1, Socket} ->
```
madninja: What's this doing here?

madninja:
See the comment re inert_read in there. In general I'd also look for other inert_read using services.. multistream_server, yamux_session, and libp2p_framed_stream. How would those function without the inert_read message being fired of somewhere?

Vagabond:
Marc said this was ok, merging.

PR #112, opened 2019-01-04
WIP Refactor peerbook process to avoid almost all gen_server calls

Vagabond:


Vagabond:
No longer WIP, tests/dialyzer passes locally (there's a pre-existing dialyzer warning).

PR #114, opened 2019-01-08
Macpie/ch1573/supervision tree checks

macpie:


macpie:
The auxiliary sup is there for stuff that should not crash the entire swarm like relay, proxy and cache.
I also fix the proxy to make the splice processes linked so that we do not leave zombie processes.
Some bug with lager and NAT.

PR #116, opened 2019-01-16
Wrap libp2p_proxy:dial_framed_stream

macpie:


src/proxy/libp2p_transport_proxy.erl
```diff
@@ -41,22 +41,23 @@ connect(_Pid, MAddr, _Options, _Timeout, TID) ->
...
+        {error, Reason} ->
+            lager:error("failed to dial proxy server ~p ~p", [PAddress, Reason]),
+            {error, fail_dial_proxy};
+        {ok, _} ->
+            receive
+                {proxy_negotiated, Socket, MultiAddr} ->
+                    Conn = libp2p_transport_tcp:new_connection(Socket, MultiAddr),
+                    lager:info("proxy successful ~p", [Conn]),
+                    libp2p_transport:start_client_session(TID, MAddr, Conn)
+            after 8000 ->
```
evanmcc: do we want this hardcoded?

madninja: yeah that doesn't seem right?

madninja: Hmmm.. on second reading I think it actually is correct.. proxying is a tcp specific feature. @macpie 
what happens if that multiaddr isn't a /ip4 multiaddr? Or is that guaranteed at this point?

macpie: it should be guaranteed at that point

src/proxy/libp2p_transport_proxy.erl
```diff
@@ -41,22 +41,37 @@ connect(_Pid, MAddr, _Options, _Timeout, TID) ->
...
+        {error, _}=Error ->
+            Error;
+        {ok, PeerInfo} ->
+            HasRelayAddress = lists:any(
+                fun libp2p_relay:is_p2p_circuit/1,
+                libp2p_peer:cleared_listen_addrs(PeerInfo)
+            ),
+            case HasRelayAddress of
+                true ->
+                    {error, invalid_proxy};
```
Vagabond: It would probably be good to log this failure case so we can tell if it happens?

PR #118, opened 2019-01-28
Nijdam/ch1808/add enacl support

madninja:
This adds support for mulitple keytypes for crypto:

* Add support for ed25519, while keeping support for ecc_compact
* Rename address to pubkey_bin to make it clearer what the thing actually is


src/libp2p_crypto.erl
```diff
@@ -2,119 +2,155 @@
...
 
--spec to_pem(private_key()) -> binary().
-to_pem(PrivKey) ->
+-spec keys_to_bin(key_map()) -> binary().
+keys_to_bin(#{secret := {ecc_compact, PrivKey}, public := {ecc_compact, _PubKey}}) ->
     PemEntry = public_key:pem_entry_encode('ECPrivateKey', PrivKey),
-    public_key:pem_encode([PemEntry]).
+    <<?KEYTYPE_ECC_COMPACT:8, (public_key:pem_encode([PemEntry]))/binary>>;
+keys_to_bin(#{secret := {ed25519, PrivKey}, public := {ed25519, PubKey}}) ->
+    io:format("PUB ~p PRIV: ~p", [byte_size(PubKey), byte_size(PrivKey)]),
```
Vagabond: Remove?

src/libp2p_crypto.erl
```diff
@@ -2,119 +2,155 @@
...
+    Bin = keys_to_bin(KeysMap),
+    file:write_file(FileName, Bin).
 
--spec to_pem(private_key()) -> binary().
-to_pem(PrivKey) ->
+-spec keys_to_bin(key_map()) -> binary().
+keys_to_bin(#{secret := {ecc_compact, PrivKey}, public := {ecc_compact, _PubKey}}) ->
     PemEntry = public_key:pem_entry_encode('ECPrivateKey', PrivKey),
-    public_key:pem_encode([PemEntry]).
+    <<?KEYTYPE_ECC_COMPACT:8, (public_key:pem_encode([PemEntry]))/binary>>;
```
Vagabond: You're mixing binary and ascii data here, this seems inadviseable?

Vagabond: Maybe it's better to just do a binary format similar to the 25519 curve?

src/libp2p_crypto.erl
```diff
@@ -2,119 +2,155 @@
...
+        _ -> false
+    end.
+
 
--spec address_to_b58(address()) -> string().
-address_to_b58(Addr) ->
-    address_to_b58(16#00, Addr).
+-spec bin_to_b58(binary()) -> string().
+bin_to_b58(Bin) ->
+    bin_to_b58(16#00, Bin).
```
Vagabond: is 16#00 the 'version'?

test/crypto_SUITE.erl
```diff
@@ -1,71 +0,0 @@
--module(crypto_SUITE).
```
Vagabond: Why was this removed?

src/libp2p_crypto.erl
Vagabond: Wondering why the 'version' is not just the key type?

src/libp2p_crypto.erl
madninja: but isn't `pem_encode(PemEntries) -> binary()` the right type signature for creating a binary here?

test/crypto_SUITE.erl
madninja: Because the SUITE was moved into the unit tests inside the .erl file a while ago. We just never noticed

src/libp2p_crypto.erl
madninja: Resolved with a87effe

src/libp2p_crypto.erl
madninja: Mostly to avoid the types and versions from overlapping.

.travis.yml
```diff
@@ -13,6 +13,7 @@ before_install:
   - cd libsodium && sudo ./configure && sudo make check && sudo make install && cd ..
 
 cache:
+  ccache: true
```
benoitc: for what do you need it? is this to speed the build of rocksdb? there is a branch that speed it. Also eventually ccache can be added to it. Let me know :)

madninja:
Yeah it’s to speed up rocksdb and libsodium builds. You have something that helps?

benoitc:
using the `static_build` branch you can run a // build using a version of cmake that supports it: 

```
ERLANG_ROCKSDB_OPTS="-DWITH_LZ4=ON -DWITH_SNAPPY=ON" ERLANG_ROCKSDB_BUILDOPTS="-j" rebar3 compile
```

will run the build in parralel with all cores with an installed version of snappy and lz4. I can add the support of `ccache` too if needed. You can also use a packaged version of rocksdb (if 0.17.2) for it using the option `WITH_SYSTEM_ROCKSDB=ON`. Let me know if it improves the build for you.

benoitc:
also with TRAVIS you can cache the built of erlang-rocksdb. This is what most do as of today.

madninja:
Right that's what I'm trying to do. The apt source list for travis is
pretty minimal so I'm not sure if rocksdb is even in there at all today,
let alone on a recent version

madninja:
@benoitc I'm going to back this out and just try to speed up libsodium builds. We can tackle the rocks speedup separately.

benoitc:
@madninja ok. fyi https://gitlab.com/barrel-db/erlang-rocksdb/commit/7cf8d66b5f1b0458891b7bdb6e1275f471db843c has support for ccache also

madninja:
Oh very nice! Can’t wait for this to merge up!

benoitc:
@madninja https://hex.pm/packages/rocksdb/1.0.0 & https://gitlab.com/barrel-db/erlang-rocksdb/blob/master/doc/customize_rocksdb_build.md ;)

PR #119, opened 2019-01-31
Return the longest matching prefix, not just the first matching prefix

Vagabond:


src/libp2p_multistream_server.erl
```diff
@@ -79,7 +79,7 @@ handle_info({inert_read, _, _}, State=#state{connection=Conn,
         "ls" ->
             handle_ls_reply(Conn, Handlers, State);
         Line ->
-            case find_handler(Line, Handlers) of
+            case find_handler(Line, Handlers, error) of
```
madninja: Is this needed?

PR #122, opened 2019-02-02
Find route to wallet from PPP addresses

fvasquez:
This PR should allow hotspots to assert location on hotspots that are connected via cellular.

The problem is that `inet_ext` is failing to parse the output from the `ip r` command as shown.

```
$ ip r
default dev ppp0 scope link 
10.59.4.14 dev ppp0 scope link 
169.254.0.0/16 dev ppp0 scope link src 169.254.172.203
```


Vagabond:
Were you able to test this?

Vagabond:
I tested this on a cellular hotspot and it worked

PR #123, opened 2019-02-07
Make it so you cannot dial yourself

macpie:


madninja:
Could you add a test ?

macpie:
ok

Vagabond:
What if you dial yourself via p2p address?

macpie:
@Vagabond it will resolve the p2p address and fail, I can add a test to show that

PR #124, opened 2019-03-06
Macpie/ch698/implement encrypted peer to peer stream for

macpie:
- [x]  Dual nonce for each way
- [x] Verify dialer/server in the peerbook

### DIFF
--- a/src/libp2p_framed_stream.erl
+++ b/src/libp2p_framed_stream.erl
@@ -224,14 +224,9 @@ init_module(Kind, Module, Connection, Args, SendPid) ->
 
 handle_info({inert_read, _, _}, #state{
                                     kind=server,
-                                    module=Module,
                                     connection=Connection,
-                                    send_pid=SendPid,
                                     secured=true,
-                                    exchanged=false,
-                                    args=Args,
-                                    pub_key=ServerPK,
-                                    priv_key=ServerSK
+                                    exchanged=false
                                 }=State0) ->
     case recv(Connection, ?RECV_TIMEOUT) of
         {error, timeout} ->
@@ -242,46 +237,19 @@ handle_info({inert_read, _, _}, #state{
             lager:notice("framed inert RECV ~p, ~p", [Error, Connection]),
             {stop, {error, Error}, State0};
         {ok, Data} ->
-            {key_exchange, ClientPK, Signature} = erlang:binary_to_term(Data),
-
-            {ok, Session} = libp2p_connection:session(Connection),
-            libp2p_session:identify(Session, self(), my_data),
-            {ok, Identify} = receive
-                {handle_identify, my_data, R} -> R
-                after 1000 -> error(timeout)
-            end,
-            PKBin = libp2p_identify:pubkey_bin(Identify),
-            ClientSwarmPK = libp2p_crypto:bin_to_pubkey(PKBin),
-
-            case libp2p_crypto:verify(erlang:term_to_binary(ClientPK), Signature, ClientSwarmPK) of
-                false ->
-                    {stop, {error, failed_verify}, State0};
-                true ->
-                    #{server_rx := RcvKey, server_tx := SendKey} = enacl:kx_server_session_keys(ServerPK, ServerSK, ClientPK),
-                    case init_module(server, Module, Connection, Args, SendPid) of
-                        {error, _}=Error ->
-                            {stop, Error, State0};
-                        {ok, State1} ->
-                            {noreply, State1#state{
-                                secured=true,
-                                exchanged=true,
-                                rcv_key=RcvKey,
-                                send_key=SendKey
-                            }}
-                    end
+            case verify_exchange(Data, State0) of
+                {error, _}=Error ->
+                    {stop, Error, State0};
+                {ok, State1} ->
+                    {noreply, State1}
             end
     end;
 handle_info({inert_read, _, _}, #state{
                                     kind=client,
-                                    module=Module,
                                     connection=Connection,
-                                    send_pid=SendPid,
                                     secured=true,
                                     parent=Parent,
-                                    exchanged=false,
-                                    args=Args,
-                                    pub_key=ClientPK,
-                                    priv_key=ClientSK
+                                    exchanged=false
                                 }=State0) ->
     case recv(Connection, ?RECV_TIMEOUT) of
         {error, timeout} ->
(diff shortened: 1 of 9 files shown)


macpie:
@madninja @Vagabond rdy for review

src/libp2p_framed_stream.erl
```diff
@@ -341,6 +493,52 @@ recv(Connection, Timeout) ->
...
+                             module=Module,
+                             connection=Connection,
+                             send_pid=SendPid,
+                             secured=true,
+                             exchanged=false,
+                             args=Args,
+                             pub_key=PK,
+                             priv_key=SK
+                       }) ->
+    {key_exchange, OtherSidePK, Signature} = erlang:binary_to_term(Data),
```
Vagabond: Might want to use the 'safe' option to binary_to_term because this is untrusted input.

Beyond that, it might be better to avoid B2T at all given portability concerns?

test/libp2p_secure_framed_stream_SUITE.erl
```diff
@@ -0,0 +1,151 @@
...
+%%--------------------------------------------------------------------
+%% TEST CASES
+%%--------------------------------------------------------------------
+
+%%--------------------------------------------------------------------
+%% @public
+%% @doc
+%% @end
+%%--------------------------------------------------------------------
+key_exchange(_Config) ->
```
Vagabond: We really need negative tests here, like where the far side is pretending to be the peer but it can't sign the ephemeral key, etc.

Vagabond:
I finally got the tests to pass, but the coverage report shows none of the new code is being run by the tests?

Vagabond:
Scratch the cover comment, I had something stale locally. Trying to figure out what happened.

macpie:
Ok Cause I used to have logs line in there to check that the encryption was called

Vagabond:
Apparently our version of rebar only writes eunit coverdata if you do `make test`. I tried upgrading rebar3 but that broke entirely.

Running eunit and ct separately does work.

Vagabond:
Is there a way to ensure or check a stream is 'secure'? For example if I'm the server, and a client dials me (for example for a consensus protocol) and I, as the server, want the stream to be secure, can I verify or guarantee that?

Vagabond:
I pushed a rebar3 upgrade to fix the test coverage thing (updated to rebar 3.9.0) you might have to delete your ~/.cache/rebar3 directory afterwards...

src/libp2p_framed_stream.erl
```diff
@@ -248,7 +400,7 @@ handle_call(Msg, From, State=#state{kind=Kind, module=Module, state=ModuleState0
                     {stop, Reason, Reply, ModuleState, Response} ->
                         {noreply, handle_resp_send({stop, Reason, From, Reply}, Response, State#state{state=ModuleState})}
                 end;
-        false -> [reply, ok, State]
```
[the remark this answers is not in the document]
macpie: yes old error fixed

Vagabond:
I pushed a test and a fix to ensure the p2p address we dial is actually the swarm on the other end.

Now, when listening server-side, we can pass `{secured, ...}` but it's not mandatory for the client to negotiate a secure connection?

macpie:
no it is not

Vagabond:
I wonder if it makes any sense to allow unsecure connections to a secure stream. I propose we require secure clients if the server lists itself as secure?

test/libp2p_secure_framed_stream_SUITE.erl
```diff
@@ -0,0 +1,236 @@
...
+        Version,
+        libp2p_secure_framed_stream_echo_test,
+        [ClientSwarm, self(), {secured, ClientSwarm}]
+    )),
+
+    ok = libp2p_swarm:stop(EvilSwarm),
+    ok = libp2p_swarm:stop(ClientSwarm),
+    ok.
+
+key_exchange(_Config) ->
```
Vagabond: So is this test for the old 'compatible' libp2p secure streams that we abandoned? If so should we remove it?

src/libp2p_framed_stream.erl
```diff
@@ -132,10 +147,42 @@ info(Pid) ->
...
 %%
-
--spec init_module(atom(), atom(), libp2p_connection:connection(), [any()]) -> {ok, #state{}} | {error, term()}.
-init_module(Kind, Module, Connection, Args) ->
+-spec pre_init(atom(), atom(), libp2p_connection:connection(), [any()], pid() | undefined) -> {ok, #state{}}
+                                                                                              | {error, term()}.
+pre_init(Kind, Module, Connection, Args, Parent) ->
     SendPid = spawn_link(libp2p_connection:mk_async_sender(self(), Connection)),
+    case proplists:get_value(secured, Args, false) of
+        Swarm when is_pid(Swarm) ->
```
Vagabond: There must be a better way to get the swarm handle here, no? maybe @madninja can advise?

Vagabond:
This appears to be working for dkg/hbbft now, and the enacl usage seems correct.

src/libp2p_framed_stream.erl
```diff
@@ -352,14 +473,35 @@ handle_fdset(State=#state{connection=Connection}) ->
...
+
+
 -spec handle_resp_send(send_result_action(), binary(), non_neg_integer(), #state{}) -> #state{}.
-handle_resp_send(Action, Data, Timeout, State=#state{sends=Sends, send_pid=SendPid}) ->
+handle_resp_send(Action, Data, Timeout, State=#state{secured=true, exchanged=true,
+                                                    send_nonce=Nonce}) ->
+    case enacl:aead_chacha20poly1305_encrypt(State#state.send_key, Nonce, <<>>, Data) of
+        {error, _Reason} ->
+            lager:warning("failed to encrypt ~p : ~p", [{Nonce, Data}, _Reason]),
+            State;
```
Vagabond: Should we close the connection if this happens?

test/libp2p_secure_framed_stream_test.erl
```diff
@@ -0,0 +1,121 @@
...
+                    #{client_rx := RcvKey, client_tx := SendKey} = enacl:kx_client_session_keys(ClientPK, ClientSK, ServerPK),
+                    {noreply, State#state{exchanged=true,
+                                          rcv_key=RcvKey,
+                                          send_key=SendKey}}
+            end
+    catch
+        _:_ ->
+            {stop, failed_bin_to_term, State}
+    end;
+handle_data(_Type, EncryptedData, #state{exchanged=true, rcv_key=RcvKey, nonce=Nonce, echo=Echo}=State) ->
```
Vagabond: shouldn't we use rcv_nonce here?

src/libp2p_framed_stream.erl
macpie: maybe?

madninja: yup, good catch

PR #125, opened 2019-03-09
Allow swarms to declare a network ID in their peerbook

Vagabond:
Network IDs are a way to prevent crosstalk between logically partitioned
libp2p networks. Swarms can declare their network ID (application
name/version, genesis block, whatever) and they will not see peer
entries from peers with a different network ID.

Swarms that do not declare a network ID can still be contacted and will
be allowed into the peerbook. Thus nodes implementing shared
infrastructure (eg. seed nodes) can be used for both sides of a logical
partition.

### DIFF
--- a/src/libp2p_swarm.erl
+++ b/src/libp2p_swarm.erl
@@ -2,6 +2,7 @@
 
 -export([start/1, start/2, stop/1, is_stopping/1, swarm/1, tid/1,
          opts/1, name/1, pubkey_bin/1, p2p_address/1, keys/1,
+         set_network_id/2, get_network_id/1,
          store_peerbook/2, peerbook/1, peerbook_pid/1, cache/1, sessions/1,
          dial/3, dial/5, connect/2, connect/4,
          dial_framed_stream/5, dial_framed_stream/7,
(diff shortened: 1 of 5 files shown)


src/peerbook/libp2p_peerbook.erl
```diff
@@ -69,6 +69,9 @@ put(#peerbook{tid=TID, stale_time=StaleTime}=Handle, PeerList) ->
                                             NewPeerId /= ThisPeerId
                                                 andalso libp2p_peer:supersedes(NewPeer, ExistingPeer)
                                                 andalso not libp2p_peer:is_stale(NewPeer, StaleTime)
+                                                andalso (libp2p_peer:network_id(NewPeer) == libp2p_swarm:get_network_id(TID)
```
madninja: I see this statement repeated but in the positive and negative form to check whether the network id matches. Could you extract this into a function on libp2p_peer to reduce copy-pasta?

Vagabond:
But it doesn't deal purely with the peer, there it already a network_id helper in peer, but this one requires some context that the peer does not contain.

madninja:
Pass in the network_id?

src/libp2p_swarm.erl
```diff
@@ -127,6 +128,21 @@ keys(TID) ->
...
     gen_server:call(Server, keys).
 
+-spec set_network_id(ets:tab() | pid(), binary()) -> ok.
+set_network_id(Sup, NetworkID) when is_pid(Sup) ->
+    set_network_id(tid(Sup), NetworkID);
+set_network_id(TID, NetworkID) ->
+    ets:insert(TID, {network_id, NetworkID}).
+
+-spec get_network_id(ets:tab() | pid()) -> binary() | undefined.
+get_network_id(Sup) when is_pid(Sup) ->
```
madninja: Could you rename this to `network_id`? makes it consistent with other getters?

Vagabond: Done

PR #126, opened 2019-03-26
Desperately search for listen addresses if we don't have any

Vagabond:


src/libp2p_transport_tcp.erl
```diff
@@ -380,30 +380,36 @@ handle_info({handle_identify, Session, {ok, Identify}}, State=#state{tid=TID}) -
...
-                                                             end
-                                                     end;
-                                                 false ->
-                                                     Acc
-                                             end
-                                     end, [], MyListenAddrs),
+            NewListenAddrsWithPid = case MyListenAddrs of
+                                        [] ->
+                                            %% engage desperation mode, we had no listen addresses before, but maybe we have some now because someone plugged in a cable or configured wifi
+                                            [ {tcp_listen_addrs(S), Pid} || [Pid, Addr, S] <- ets:match(TID, {{libp2p_config:listen_socket(), '$1'}, {'$2', '$3'}}), match_addr(Addr, TID) /= false ];
```
madninja: Pull the ets lookup into libp2p_config so we avoid exposing internals?

PR #129, opened 2019-04-06
quieter libp2p + some elections-related cleanup

evanmcc:
- moves some stuff to debug
- makes sure that elections don't cause infinite disk usage
- allows options to get to relcast


Vagabond:
lgtm

PR #133, opened 2019-04-10
Avoid crashes on dial back

macpie:


src/proxy/libp2p_stream_proxy.erl
```diff
@@ -117,8 +117,12 @@ handle_server_data({req, Req}, Env, #state{swarm=Swarm}=State) ->
     end;
 handle_server_data({dial_back, DialBack}, Env, State) ->
     lager:info("server got dial back request ~p", [DialBack]),
-    {ok, Connection} = dial_back(Env, State),
-    {noreply, State#state{raw_connection=Connection}};
+    case dial_back(Env, State) of
+        {ok, Connection} ->
+            {noreply, State#state{raw_connection=Connection}};
+        {error, Reason} ->
```
evanmcc: this is still going to crash.  we need to filter out common errors like timeouts and tcp closed and then `{stop, normal, State}` on those.

PR #139, opened 2019-04-23
Fixes for relay/proxy

Vagabond:
Prior to this change we have 2 issues:

* Nodes that did not need to advertise a relay listen address were doing so
* We refused to dial through a host with a relay address if that host itself was listed as a relay for another node

Fixing the first issue is not enough, as the second restriction is too broad. So we remove any relay stuff if our NAT type changes such that we know we don't need it AND we also add a `no_relay` dial flag so that if we're trying to dial a relay/proxy, we ignore the relay/proxy transport for dialing the intermediate host.

Given the internet's topology it's very unlikely we need to hop through multiple relays to get to a distant host, so this should solve most issues.

### DIFF
--- a/src/libp2p_transport_tcp.erl
+++ b/src/libp2p_transport_tcp.erl
@@ -471,7 +471,8 @@ handle_info({stungun_nat, TxnID, NatType}, State=#state{tid=TID, stun_txns=StunT
             lager:debug("stungun detected NAT type ~p", [NatType]),
             case NatType of
                 none ->
-                    %% don't need to start relay here
+                    %% don't need to start relay here, stop any we already have
+                    libp2p_relay_server:stop(libp2p_swarm:swarm(TID)),
                     %% if we didn't have an external address originally, set the NAT type to 'static'
                     case State#state.negotiated_nat of
                         true ->
--- a/src/proxy/libp2p_stream_proxy.erl
+++ b/src/proxy/libp2p_stream_proxy.erl
@@ -35,6 +35,7 @@
     id :: binary() | undefined,
     transport :: pid() | undefined,
     swarm :: pid() | undefined,
+    proxy_address :: undefined | string(),
     connection :: libp2p_connection:connection(),
     raw_connection :: libp2p_connection:connection() | undefined
 }).
@@ -60,13 +61,14 @@ init(server, Conn, [_, _Pid, TID]=Args) ->
 init(client, Conn, Args) ->
     lager:info("init proxy client with ~p", [{Conn, Args}]),
     ID = proplists:get_value(id, Args),
+    Swarm = proplists:get_value(swarm, Args),
     case proplists:get_value(p2p_circuit, Args) of
         undefined ->
             {ok, #state{id=ID, connection=Conn}};
         P2PCircuit ->
             TransportPid = proplists:get_value(transport, Args),
             self() ! {proxy_req_send, P2PCircuit},
-            {ok, #state{id=ID, transport=TransportPid, connection=Conn}}
+            {ok, #state{swarm=Swarm, id=ID, transport=TransportPid, connection=Conn}}
 
     end.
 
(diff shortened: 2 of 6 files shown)


src/libp2p_transport_tcp.erl
```diff
@@ -471,7 +471,8 @@ handle_info({stungun_nat, TxnID, NatType}, State=#state{tid=TID, stun_txns=StunT
             lager:debug("stungun detected NAT type ~p", [NatType]),
             case NatType of
                 none ->
-                    %% don't need to start relay here
+                    %% don't need to start relay here, stop any we already have
+                    libp2p_relay_server:stop(libp2p_swarm:swarm(TID)),
```
macpie: if this server was started then stopped should we remove relay address from peerbook?

src/proxy/libp2p_transport_proxy.erl
```diff
@@ -44,34 +44,19 @@ connect(_Pid, MAddr, _Options, _Timeout, TID) ->
...
-                                {proxy_negotiated, Socket, MultiAddr} ->
-                                    Conn = libp2p_transport_tcp:new_connection(Socket, MultiAddr),
-                                    lager:info("proxy successful ~p", [Conn]),
-                                    libp2p_transport:start_client_session(TID, MAddr, Conn)
-                            after 8000 ->
-                                {error, timeout_relay_session}
-                            end
-                    end
+           ],
+    case libp2p_proxy:dial_framed_stream(Swarm, PAddress, Args) of
```
macpie: Should this be forced to a non relay address?

src/libp2p_transport_tcp.erl
Vagabond: The terminate for the relay server handles that

macpie: yep I missed the `true = libp2p_config:remove_relay(TID),`

src/relay/libp2p_relay_server.erl
```diff
@@ -66,6 +68,18 @@ connection_lost(Swarm) ->
             Error
     end.
 
+-spec stop(pid()) -> ok | {error, any()}.
+stop(Swarm) ->
+    case get_relay_server(Swarm) of
+        {ok, Pid} ->
+            %% it's a permanant worker, so just stop it and let it restart
+            %% without having relay() called on it
```
madninja: Why not make relay server a transient instead?

src/proxy/libp2p_transport_proxy.erl
Vagabond: This should be unreachable because it's only called from relay:connect, which checks for the no_relay flag

src/relay/libp2p_relay_server.erl
Vagabond: Because I didn't want to do too much surgery on this without knowing if it will help, I guess

Vagabond: It currently always runs, and we just conditionally tell it to start maintaining a relay address for us. I just stop it to return it to the initial state after it restarts.

Vagabond:
Ok, while testing proxy I discovered that there was a bug hiding behind the blanket restriction on dialing through peers that listed a relay address in their peerbook:

We were using the ephemeral port of the 'raw connection' rather than the peer's listen port to dial back, so it would fail. This final commit changes that so we look up the proxy host's address in the peerbook and match the listen address by IP to find the listen port and dial using that.

src/proxy/libp2p_stream_proxy.erl
```diff
@@ -35,6 +35,7 @@
     id :: binary() | undefined,
     transport :: pid() | undefined,
     swarm :: pid() | undefined,
+    proxy_address :: any(),
```
madninja: When is this more than a `string()` ?

madninja: Or undefined I guess

Vagabond: I forgot what type it was

PR #140, opened 2019-04-24
support for various relcast API changes

evanmcc:
mostly in support of better ack batching, message batching, and sync delay in general.

### DIFF
--- a/src/group/libp2p_ack_stream.erl
+++ b/src/group/libp2p_ack_stream.erl
@@ -63,6 +63,7 @@ handle_data(_Kind, Data, State=#state{ack_ref=AckRef, ack_module=AckModule, ack_
             AckModule:handle_ack(AckState, AckRef, Seq, Reset == true),
             {noreply, State};
         _Other ->
+            lager:info("other ~p", [_Other]),
             {noreply, State}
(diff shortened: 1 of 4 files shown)


src/group/libp2p_ack_stream.erl
```diff
@@ -53,23 +53,24 @@ init(client, Connection, [AckRef, AckModule, AckState]) ->
 
 handle_data(_Kind, Data, State=#state{ack_ref=AckRef, ack_module=AckModule, ack_state=AckState}) ->
     case libp2p_ack_stream_pb:decode_msg(Data, libp2p_ack_frame_pb) of
-        #libp2p_ack_frame_pb{data=Bin, seq=[Seq]} when Bin /= <<>> ->
+        #libp2p_ack_frame_pb{messages=[Bin], seqs=[Seq]} when Bin /= <<>> ->
```
Vagabond: Presumably these becoming a list are so we can do batching next?

evanmcc: yes.

Vagabond:
This has a known dialyzer error, but we can't figure it out for now.

PR #141, opened 2019-04-24
Fix a socket/process leak in proxy

Vagabond:
Before this change, we left a gen_server hanging around on the proxy server.
Unfortunately, this was the process that ranch thought owned the socket.
So, once the socket closed, the gen_server remained alive and so the
number of sockets ranch thought were open never decremented. This change
ensures that the socket owners are cleaned up and a test is extended to
verify this.

### DIFF
--- a/src/proxy/libp2p_proxy_session.erl
+++ b/src/proxy/libp2p_proxy_session.erl
@@ -5,87 +5,23 @@
 %%%-------------------------------------------------------------------
 -module(libp2p_proxy_session).
 
--behavior(gen_server).
-
 %% ------------------------------------------------------------------
 %% API Function Exports
 %% ------------------------------------------------------------------
 -export([
-    start_server/4,
-    start_link/1
+    start_server/4
 ]).
 
-%% ------------------------------------------------------------------
-%% gen_server Function Exports
-%% ------------------------------------------------------------------
--export([
-    init/1,
-    handle_call/3,
-    handle_cast/2,
-    handle_info/2,
-    terminate/2,
-    code_change/3
-]).
-
--record(state, {
-    type :: client | server,
-    connection :: libp2p_connection:connection(),
-    path :: string(),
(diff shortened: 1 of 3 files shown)


src/proxy/libp2p_proxy_session.erl
```diff
@@ -72,8 +72,10 @@ handle_info(transfer_socket, #state{type=server, connection=Connection, path=Pat
     lager:info("doing socket transfer"),
     <<"/", ID0/binary>> = erlang:list_to_binary(Path),
     ID1 = base58:base58_to_binary(erlang:binary_to_list(ID0)),
+    %% this will block until the connection is finished with
     ok = libp2p_proxy_server:connection(TID, Connection, ID1),
-    {noreply, State};
+    %% we did our job, time to die
```
evanmcc: do we actually need a gen_server here?

test/proxy_SUITE.erl
```diff
@@ -132,10 +132,23 @@ basic(_Config) ->
...
+
     ok = libp2p_swarm:stop(ServerSwarm),
-    ok = libp2p_swarm:stop(ProxySwarm),
     ok = libp2p_swarm:stop(ClientSwarm),
 
+    timer:sleep(5000),
+
+    %% check we didn't leak any sockets here
+    [{_ID, Info}] = ranch:info(),
+    ?assertEqual(0, proplists:get_value(active_connections, Info)),
```
evanmcc: wait until?

Vagabond:
The process dying tells ranch we're done with it. I tried the 'remove_socket' stuff but it didn't seem to work as I expected.

src/proxy/libp2p_proxy_session.erl
Vagabond: Probably not

Vagabond:
I have made the requested updates

PR #142, opened 2019-04-26
Fix 2 bugs in relay sorting so host with the least connections is first

Vagabond:
We had a typo, and we were sorting backwards (lower should be first)


Vagabond:
The dialyzer error will be fixed once we land batching in relcast

PR #143, opened 2019-04-26
relcast batching support for libp2p

evanmcc:
relies on: https://github.com/helium/relcast/pull/24

### DIFF
--- a/src/group/libp2p_group_relcast_server.erl
+++ b/src/group/libp2p_group_relcast_server.erl
@@ -19,7 +19,6 @@
          in_flight = 0 :: non_neg_integer(),
          connects = 0 :: non_neg_integer(),
          last_take = unknown :: atom(),
-         last_in_seq :: undefined | pos_integer(),
          last_ack = 0 :: non_neg_integer()
        }).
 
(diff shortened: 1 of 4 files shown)


src/group/libp2p_group_relcast_server.erl
```diff
@@ -247,8 +245,27 @@ handle_cast({handle_data, Index, Msg, Seq}, State=#state{self_index=_SelfIndex})
...
+             ({Seq, Msg}, RC) ->
+                  case relcast:deliver(Seq, Msg, Index, RC) of
+                      full ->
+                          {full, RC, [{Seq, Msg}]};
+
+                  {ok, NewRC} ->
+                      NewRC;
+                  %% just keep looping, I guess, it kind of doesn't matter?
+                  {stop, Timeout, NewRC} ->
+                      erlang:send_after(Timeout, self(), force_close),
```
Vagabond: should we limit the number of timers we set here?

evanmcc: we could, but subsequent timers won't have any further effect since after the first one lands, the process will be dead and the message will be dropped by ERTS.

Vagabond: Yeah.

PR #144, opened 2019-04-26
Add stream_type to a framed_streams process dictionary

madninja:
It maeks it easier to detect what stream is running


madninja:
Known build failure.. tests pass locally.. Merging while @evanmcc fixes the typecheck errors

PR #145, opened 2019-04-26
Propagate error from proxy server to client

macpie:


Vagabond:
Can you write up what this actually *does* when the failures make it back to each side?

src/proxy/libp2p_proxy_server.erl
```diff
@@ -99,8 +99,23 @@ handle_call({init_proxy, ID, ServerStream, SAddress}, _From, #state{data=Data,
...
+handle_call({init_proxy, ID, _ServerStream, SAddress}, _From, #state{swarm=Swarm}=State) ->
     lager:warning("cannot process proxy request for ~p, ~p. Limit exceeded", [ID, SAddress]),
+    erlang:spawn(
+        fun() ->
+            case libp2p_proxy:dial_framed_stream(Swarm, SAddress, [{id, ID}, {swarm, Swarm}]) of
+                {ok, ClientStream} ->
+                    Ref = erlang:monitor(process, ClientStream),
+                    ClientStream ! proxy_overloaded,
+                    receive
+                        {'DOWN', Ref, process, ClientStream, _} ->
```
Vagabond: Should we have a timeout here, maybe one that would kill the clientstream process?

src/proxy/libp2p_stream_proxy.erl
```diff
@@ -143,6 +150,14 @@ handle_server_data({resp, Resp}, _Env, #state{swarm=Swarm, raw_connection=Connec
     %% So this process apparently can't die, at least not yet
     %% instead remember the monitor ref for later
     {noreply, State#state{connection_ref=MRef}};
+handle_server_data({overload, Overload}, _Env, #state{swarm=Swarm}=State) ->
+    PubKeyBin = libp2p_proxy_overload:pub_key_bin(Overload),
+    R = libp2p_crypto:pubkey_bin_to_p2p(PubKeyBin),
+    A = libp2p_swarm:p2p_address(Swarm),
+    P2PCircuit = libp2p_relay:p2p_circuit(R, A),
+    TID = libp2p_swarm:tid(Swarm),
+    true = libp2p_config:remove_listener(TID, P2PCircuit),
```
Vagabond: Do we need to signal the relay server to obtain a new relay connection here?

macpie: Yes potentially but not sure how to avoid them picking the same one again?

Vagabond: The relay server already has some logic to handle that, at least on-average.

PR #148, opened 2019-04-29
Add empty timeout to yamux session and streams

madninja:


src/libp2p_yamux_session.erl
```diff
@@ -6,6 +6,7 @@
 
 -define(TIMEOUT, 5000).
 -define(LIVENESS_TIMEOUT, 30000).
+-define(DEFAULT_IDLE_TIMEOUT, 10000).
```
Vagabond: I think this should be 30 seconds

src/libp2p_yamux_stream.erl
```diff
@@ -48,6 +50,7 @@
 -define(WAITER_DATA(S), S#state.recv_state#recv_state.waiter_data).
 -define(RECEIVABLE_SIZE(S), (byte_size(?WINDOW_DATA(S)) + byte_size(?WAITER_DATA(S)))).
 -define(REMOTE_CLOSED(S), ?CLOSE_STATE(S) == pending).
+-define(DEFAULT_IDLE_TIMEOUT, 5000).
```
Vagabond: I would make this 15 seconds, or so?

PR #159, opened 2019-06-13
Handle server switch in relay dial_back

macpie:


src/relay/libp2p_relay_server.erl
```diff
@@ -86,7 +86,19 @@ init(TID) ->
     true = libp2p_config:insert_relay(TID, self()),
     {ok, #state{tid=TID, swarm=Swarm}}.
 
-handle_call(init_relay, _From, #state{started=false, swarm=Swarm}=State0) ->
+handle_call(init_relay, _From, State) ->
```
Vagabond: This should move too?

macpie: fixed

PR #163, opened 2019-06-26
Pd/nat server

macpie:


src/nat/libp2p_nat.erl
```diff
@@ -61,30 +61,7 @@ spawn_discovery(Pid, MultiAddrs, TID) ->
...
-                {ok, ExtAddr, ExtPort, Lease, Since} ->
-                    _ = libp2p_nat_statem:register(Statem, ExtPort, Lease, Since),
-                    {ok, ParsedExtAddress} = inet_parse:address(ExtAddr),
-                    ExtMultiAddr = libp2p_transport_tcp:to_multiaddr({ParsedExtAddress, ExtPort}),
-                    Pid ! {nat_discovered, MultiAddr, ExtMultiAddr},
-                    ok;
-                {error, _Reason1} ->
-                    lager:warning("unable to add nat mapping: ~p", [_Reason1])
-            end
+            {ok, _Server} = libp2p_nat_server:start([Pid, TID, MultiAddr, Port])
```
Vagabond: Return spec says it should be `ok`.

PR #164, opened 2019-06-26
blow away the cache less often, don't validate peers on folds

evanmcc:
on the spots, we're seeing this take quite a long time even with smaller peerbooks.  I don't think we need to do full validation each time.

### DIFF
--- a/src/peerbook/libp2p_peer.erl
+++ b/src/peerbook/libp2p_peer.erl
@@ -16,7 +16,7 @@
 -type metadata() :: [{string(), binary()}].
 -export_type([peer/0, association/0, peer_map/0, nat_type/0]).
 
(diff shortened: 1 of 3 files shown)


Vagabond:
This looks fine to me, this performance issue was in code that I think I wrote originally...

src/peerbook/libp2p_peer.erl
```diff
@@ -332,6 +332,10 @@ decode(Bin) ->
     verify(Msg),
     Msg.
 
+%% @doc Decodes a binary peer without verification, use with care.
+-spec decode_lite(binary()) -> peer().
+decode_lite(Bin) ->
```
madninja: Could you rename this decode_unsafe? And have the other decode call it before verifying?

Vagabond: I will fix this as Evan is in transit.

PR #170, opened 2019-07-02
Handle case when connection we're identifying gets closed

Vagabond:
This simply adds a try/catch

### DIFF
--- a/src/libp2p_transport_tcp.erl
+++ b/src/libp2p_transport_tcp.erl
@@ -386,72 +386,14 @@ handle_info({handle_identify, Session, {error, Error}}, State=#state{}) ->
     {_LocalAddr, PeerAddr} = libp2p_session:addr_info(Session),
     lager:notice("session identification failed for ~p: ~p", [PeerAddr, Error]),
     {noreply, State};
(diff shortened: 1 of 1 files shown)


src/libp2p_transport_tcp.erl
```diff
@@ -387,66 +387,72 @@ handle_info({handle_identify, Session, {error, Error}}, State=#state{}) ->
...
+                        false ->
+                            lager:debug("identify response with local address ~p that is not a listen addr socket ~p, ignoring",
+                                        [LocalAddr, NewListenAddrs]),
+                            %% this is likely a discovery session we dialed with unique_port
+                            %% we can't trust this for the purposes of the observed address
+                            {noreply, State}
+                    end
             end
+    catch
+        _:_ ->
```
evanmcc: should we be logging things here so we can more selectively filter later?

madninja: or maybe a (case catch) to keep it in the case statements and reduce one indent level in that insanely large code block?

madninja: I realize there are a couple of places an error can be thrown so the top level try block may be the best way to do it, but this section of code is getting pretty unwieldy, no?

Vagabond:
Comments should be addressed?

PR #175, opened 2019-07-05
Professor/ch3819/external ipv4 addresses should sort before

macpie:


src/libp2p_transport.erl
```diff
@@ -41,21 +41,29 @@ for_addr(TID, Addr) ->
...
+        libp2p_config:lookup_transports(TID)
     ),
-
-    case lists:sort(Matches) of
+    case Matches of
         [] ->
             {error, {unsupported_address, Addr}};
-        [{_, Matched, {Transport, Pid}}|_] ->
-            {ok, Matched, {Transport, Pid}}
+        [{Matched, {Transport, Pid}}] ->
```
macpie: We should only match one transport

macpie:
Each transport can now, prioritize each individual address. This allow tcp to make the difference for `rfc1918` IPs.

src/proxy/libp2p_transport_proxy.erl
```diff
@@ -28,12 +27,9 @@ start_listener(_Pid, _Addr) ->
...
     false.
 
--spec sort_addrs([string()]) -> [string()].
+-spec sort_addrs([string()]) -> [{integer(), string()}].
 sort_addrs(Addrs) ->
-    Addrs.
-
--spec priority() -> integer().
-priority() -> 99.
+    [{5, A} || A <- Addrs].
```
Vagabond: Why not keep this as 99?

src/libp2p_transport.erl
```diff
@@ -41,21 +41,29 @@ for_addr(TID, Addr) ->
...
+            {ok, Matched, {Transport, Pid}};
+        [{_Matched, {_Transport, _Pid}}|_] ->
+            {error, {multiple_match_address, Addr}}
     end.
 
+%%--------------------------------------------------------------------
+%% @doc
+%% Addresses are sorted by priority randing from 1 to 5 (1 highest and 5 lowest priority)
```
Vagabond: should be `ranking`

src/proxy/libp2p_transport_proxy.erl
macpie: idk just wanted a smaller range  99 seemed just useless

Vagabond: I guess this will never pass `match_addr` so it's academic

macpie: yes

Vagabond:
This looks good, let's give it a try in the field.

PR #177, opened 2019-07-08
Macpie/ch3861/stale nat hole not removed from peerbook

macpie:


src/relay/libp2p_relay_server.erl
```diff
@@ -14,7 +14,7 @@
     start_link/1,
     relay/1,
     stop/1,
-    connection_lost/1
+    negociated/2
```
madninja: typo. it's `negotiated`

src/relay/libp2p_relay_server.erl
```diff
@@ -105,73 +115,73 @@ handle_cast(init_relay, #state{started=false, swarm=Swarm}=State0) ->
...
-            {noreply, add_flap(State#state{started=true, connection=Connection})};
-        {error, no_peer} ->
-            lager:warning("could not initiate relay no peer found"),
-            {noreply, State};
-        {error, no_address} ->
-            lager:warning("could not initiate relay, swarm has no listen address"),
-            {noreply, State};
+handle_info({new_peers, NewPeers}, #state{tid=TID, stream=Pid}=State) when is_pid(Pid) ->
+    Swarm = libp2p_swarm:swarm(TID),
+    SwarmAddr = libp2p_swarm:pubkey_bin(Swarm),
```
madninja: Technically it's not the SwarmAddr, but the SwarmPubKey? No big deal but the name got me confused with it's p2p address for a sec

src/relay/libp2p_relay_server.erl
```diff
@@ -105,73 +115,73 @@ handle_cast(init_relay, #state{started=false, swarm=Swarm}=State0) ->
...
+    ok;
+terminate(_Reason, _State) ->
     ok.
 
 %% ------------------------------------------------------------------
 %% Internal Function Definitions
 %% ------------------------------------------------------------------
+
+retry(#state{retrying=undefined}=State) ->
+    TimeRef = erlang:send_after(2500, self(), retry),
```
madninja: Should we use the `backoff` library here like we do in group_worker to back of retrying too regularly?

macpie: we use to have an issue with this but it should be fixed now so I don't think backoff really matters

madninja: so if we don't have an issue do we still want to retry?

src/relay/libp2p_relay_server.erl
```diff
@@ -81,97 +80,108 @@ stop(Swarm) ->
...
 %% ------------------------------------------------------------------
 init(TID) ->
-    lager:debug("~p init with ~p", [?MODULE, TID]),
-    Swarm = libp2p_swarm:swarm(TID),
+    lager:info("~p init with ~p", [?MODULE, TID]),
     true = libp2p_config:insert_relay(TID, self()),
-    {ok, #state{tid=TID, swarm=Swarm}}.
+    {ok, #state{tid=TID}}.
 
+handle_call({negotiated, Address, Pid}, _From, #state{tid=TID, stream=Pid}=State) when is_pid(Pid) ->
```
Vagabond: Does this need to be a call?

Vagabond: oh, I see there's other clauses of this, nevermind.

src/relay/libp2p_relay_server.erl
macpie: the issue was we were retrying all the time sending A LOT of messages. I don't really want to back off here as I want to get an address asap

PR #178, opened 2019-07-12
Do not take down stream and remove listener

macpie:


Vagabond:
I think the other half of this is to obtain a *new* circuit address via the relay server?

madninja:
@macpie I think the log is pointing out some valid errors here

macpie:
what do you mean?

madninja:
> what do you mean?

Check out the build logs in buildkite (clicks the Details link above). There are a few errors in there and I think at least one of those is a valid issue, not some transient network error

macpie:
seems like buildkite is passing now

PR #180, opened 2019-07-15
When dialing a p2p address, report the error for each address

Vagabond:


madninja:
@Vagabond can you fix the error in swarm_SUITE that shows up in the build logs?

PR #181, opened 2019-07-15
Retry dial with reuseport disabled if eaddrnotavail is encountered

Vagabond:


src/libp2p_transport_tcp.erl
```diff
@@ -591,6 +591,8 @@ connect_to(Addr, UserOptions, Timeout, TID, TCPPid) ->
                             {ok, SessionPid};
                         {error, Reason} -> {error, Reason}
                     end;
+                {error, eaddrnotavail} when UniquePort == false ->
+                    connect_to(Addr, [unique_port | UserOptions], Timeout, TID, TCPPid);
```
madninja: Can you use the tuple version of this please? `{unique_port, true}` .
Also add a comment here since this changes the behavior of libp2p_swarm:connect a bit

PR #190, opened 2019-07-25
Tweaks to relay and proxy

Vagabond:
* Time out spliced sockets due to inactivity
* Blacklist relay peers we've explicitly disconnected from for a period of time

### DIFF
--- a/rebar.lock
+++ b/rebar.lock
@@ -38,7 +38,7 @@
  {<<"small_ints">>,{pkg,<<"small_ints">>,<<"0.1.0">>},1},
  {<<"splicer">>,
   {git,"https://github.com/helium/erlang-splicer.git",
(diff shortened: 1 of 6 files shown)


Vagabond:
> This mostly looks good; I'm wondering about two things, though:
> 
>     * are there any cases where we disconnect from a relay peer 'without prejudice' as it were, where we don't want avoid using them for the ban timer?

Yes, I've adjusted the code to take this into account.

> 
>     * should we have a fallback when there are no usable peers but there are some on the banlist we could retry?

Peers get banned for 5 minutes, and we only ban up to the last 1/3 of the peers we know about now.

Vagabond:
This has a failing test, but I'm merging it for expediency.

PR #191, opened 2019-07-26
Rename relay_server:stop/2 to replace to better reflact intent and usage

Vagabond:
Additionally, ensure that a relay replace triggers a relay retry and
that we apply the banlist before triggering the retry.

### DIFF
--- a/src/relay/libp2p_relay_server.erl
+++ b/src/relay/libp2p_relay_server.erl
@@ -13,7 +13,7 @@
 -export([
     start_link/1,
     relay/1,
-    stop/1, stop/2,
+    stop/1, replace/2,
     negotiated/2
 ]).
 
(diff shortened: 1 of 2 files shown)


src/relay/libp2p_relay_server.erl
```diff
@@ -173,15 +173,15 @@ handle_info({'DOWN', _Ref, process, Pid, Reason}, #state{tid=TID, stream=Pid, ad
...
+    {noreply, retry(banlist(Address, State#state{stream=undefined, address=undefined}))};
+handle_info({replace_relay, Address}, #state{stream=Pid, address=Address, tid=TID}=State) when is_pid(Pid) ->
+    lager:warning("relay was asked to be replaced ~p ~p", [Pid, Address]),
     _ = libp2p_config:remove_listener(TID, Address),
     catch libp2p_framed_stream:close(Pid),
     %% we likely disconnected from this relay specifically because of overload or max duration
     %% so avoid immediately re-connecting to it
-    {noreply, banlist(Address, State#state{stream=undefined, address=undefined})};
-handle_info({stop_relay, _OtherAddress}, State) ->
+    {noreply, retry(banlist(Address, State#state{stream=undefined, address=undefined}))};
```
evanmcc: a call to retry was added here? intentional?

Vagabond: Yes, because we want to obtain a new relay address.

PR #192, opened 2019-07-27
Macpie/stale

macpie:


src/peerbook/libp2p_peerbook.erl
```diff
@@ -306,7 +306,8 @@ mk_this_peer(CurrentPeer, State=#state{tid=TID}) ->
                             connected => ConnectedAddrs,
                             nat_type => State#state.nat_type,
                             network_id => NetworkID,
-                            associations => Associations},
+                            associations => Associations,
+                            timestamp => erlang:system_time(millisecond) + StaleTime},
```
macpie: my goal here is: to set this in the future so that peers do not look stale all the time. Otherwise they are set to `erlang:system_time(millisecond)` by default in `libp2p_peer:from_map`.

macpie: I have no idea if I found a bug or I am just miss understanding the `stale_time` stuff

src/relay/libp2p_relay_server.erl
```diff
@@ -269,8 +269,9 @@ shuffle(List) ->
 
 %% merge new peers into old peers based on their address
 merge_peers(NewPeers, OldPeers) ->
-    maps:values(maps:merge(maps:from_list([{libp2p_peer:pubkey_bin(P), P} || P <- OldPeers]),
-                           maps:from_list([{libp2p_peer:pubkey_bin(P), P} || P <- NewPeers]))).
+    Peers = maps:values(maps:merge(maps:from_list([{libp2p_peer:pubkey_bin(P), P} || P <- OldPeers]),
```
macpie: This is the end goal to remove stale peers from relay_server, because we see a lot of them (relay server) having 600+ peers in state, which normally should be impossible but somehow happening so my guess is there are a lot of stale peers in this state.

Last resort would be to not use notify and just grab peers from peerbook everytime

src/peerbook/libp2p_peerbook.erl
macpie: forget this I did misunderstand stale time

macpie:
Tested with 
```
{ok, Pid} = libp2p_config:lookup_relay(libp2p_swarm:tid(blockchain_swarm:swarm())), P1 = element(3, sys:get_state(Pid)),
{length(P1), length(lists:filter(fun(P) -> not libp2p_peer:is_stale(P, timer:hours(24)) end, P1))}.
``` 

looks like that should fix or 600 peers issue 👍

src/peerbook/libp2p_peer.erl
```diff
@@ -8,7 +8,8 @@
                        connected => [binary()],
                        nat_type => nat_type(),
                        network_id => binary(),
-                       associations => association_map()
+                       associations => association_map(),
+                       timestamp => non_neg_integer()
```
madninja: Why is the timestamp for a peer here? The stale time is only for the peer of _this_ swarm, meaning the peerbook manages it

test/peer_SUITE.erl
```diff
@@ -156,3 +157,18 @@ association_test(_) ->
...
+
+stale_test(_) ->
+    #{secret := PrivKey1, public := PubKey1} = libp2p_crypto:generate_keys(ecc_compact),
+    SigFun1 = libp2p_crypto:mk_sig_fun(PrivKey1),
+
+    PeerMap = #{pubkey => libp2p_crypto:pubkey_to_bin(PubKey1),
+                listen_addrs => ["/ip4/8.8.8.8/tcp/1234"],
+                connected => [],
+                nat_type => static,
+                timestamp => erlang:system_time(millisecond) + 100},
```
madninja: what's this timestamp actually do?

src/peerbook/libp2p_peer.erl
macpie: This was before I understood what timestamp and stale_time were doing

test/peer_SUITE.erl
macpie: This was just a test out of curiosity will remove

Vagabond:
Are we good on this?

macpie:
I think we are? @madninja

PR #194, opened 2019-07-31
Do not guess session for relay

macpie:


src/relay/libp2p_stream_relay.erl
```diff
@@ -208,14 +208,18 @@ handle_server_data({bridge_cr, Bridge}, _Env, #state{swarm=Swarm}=State) ->
...
+handle_server_data({bridge_sc, Bridge}, _Env,#state{swarm=Swarm, connection=Conn}=State) ->
     Client = libp2p_relay_bridge:client(Bridge),
     Server = libp2p_relay_bridge:server(Bridge),
     lager:debug("Client (~s) got Server (~s) dialing back", [Client, Server]),
-    SessionPids = [Pid || {_, Pid} <- libp2p_swarm:sessions(Swarm)],
     case libp2p_config:lookup_relay_sessions(libp2p_swarm:tid(Swarm), Server) of
-        false -> ok;
-        {ok, Pid} -> Pid ! {sessions, SessionPids}
+        false ->
+            ok;
```
evanmcc: can you add a message back to the other pid here so it can fail without waiting 15s?

src/relay/libp2p_stream_relay.erl
```diff
@@ -208,14 +208,18 @@ handle_server_data({bridge_cr, Bridge}, _Env, #state{swarm=Swarm}=State) ->
...
+        {ok, Pid} ->
+            case libp2p_connection:session(Conn) of
+                {error, _} -> ok;
```
evanmcc: also here

src/relay/libp2p_transport_relay.erl
```diff
@@ -130,10 +130,10 @@ check_peerbook(TID, MAddr) ->
 -spec connect_rcv(pid(), string(), string(), pid()) -> {ok, pid()} | {error, term()}.
 connect_rcv(Swarm, MAddr, SAddress, SessionPid) ->
     receive
```
evanmcc: do we know the pid of the process that we're expecting a message from here?  if possible, can we monitor is so that if it's dead we'll know right away instead of blocking for 15s?

src/relay/libp2p_stream_relay.erl
macpie: here we can't cause I don't have the pid

src/relay/libp2p_stream_relay.erl
macpie: But I can do it here

src/relay/libp2p_transport_relay.erl
```diff
@@ -143,7 +144,10 @@ connect_rcv(Swarm, MAddr, SAddress, SessionPid) ->
         {error, _Reason}=Error ->
             true = libp2p_config:remove_relay_sessions(libp2p_swarm:tid(Swarm), SAddress),
             lager:error("no relay sessions ~p", [_Reason]),
-            Error
+            Error;
+        {'DOWN', _Ref, process, _Stream, _Reason} ->
```
Vagabond: You should pass in the stream pid or the monitor ref here so you can match on them rather than just a random DOWN message.

Vagabond:
So the proxy transport has a very similar pattern with `connect_rcv`, we should apply some of these improvements over there too.

src/relay/libp2p_stream_relay.erl
evanmcc: why not?

Vagabond: Because this is a failure to look up the pid we're trying to message.

src/relay/libp2p_transport_relay.erl
```diff
@@ -86,12 +86,13 @@ connect_to(_Pid, MAddr, Options, Timeout, TID) ->
             Error;
         {ok, SessionPid} ->
             Swarm = libp2p_swarm:swarm(TID),
-            {ok, _} = libp2p_relay:dial_framed_stream(
+            {ok, Stream} = libp2p_relay:dial_framed_stream(
                 Swarm,
                 RAddress,
                 [{type, {bridge_cr, MAddr}}]
```
Vagabond: can you pass self() here and avoid having to lookup the pid to message back?

evanmcc: I think that you can do it without passing.  This goes through several other `dial_framed_stream` calls, but the last one could call self() before spawning.

evanmcc: I think this is the call at the bottom and we're still in the dialing process: https://github.com/helium/erlang-libp2p/blob/master/src/libp2p_session.erl#L75

I just suggest this because it might be an easier change than threading self through all the intervening layers.

Vagabond: You can also just prefix/suffix it to the parameter list there and it will get passed through...

Vagabond: I looked into doing this, and it seems quite a bit more complicated. Propose we skip it for now.

PR #195, opened 2019-08-06
try harder to not leak processes and ranch connections

evanmcc:


src/proxy/libp2p_proxy_server.erl
```diff
@@ -239,18 +239,27 @@ dial_back(ID, ServerStream, ClientStream) ->
...
+                    try
+                        receive control_given -> ok
+                        after 10000 -> throw(control_transfer_timeout)
+                        end,
+                        Socket1 = libp2p_connection:socket(Connection1),
+                        {ok, FD1} = inet:getfd(Socket1),
+                        Socket2 = libp2p_connection:socket(Connection2),
+                        {ok, FD2} = inet:getfd(Socket2),
+                        splicer:splice(FD1, FD2, ?SPLICE_TIMEOUT),
+                        (catch libp2p_connection:close(Connection1)),
```
Vagabond: should the connection closes be in the `after` stanza so they always run?

evanmcc: we don't actually need an after stanza, I think.  I just moved those and the after stuff after the `end` of the `try` expression.

PR #196, opened 2019-08-07
Close and terminate proxy connections that timed out before splicing

Vagabond:


macpie:
codev failed to upload data again... Otherwise tests are passing

PR #197, opened 2019-08-08
Add new swarm option to attached signed metadata to the swarm's peer entry

Vagabond:


test/peer_SUITE.erl
```diff
@@ -155,4 +156,19 @@ association_test(_) ->
...
+    #{secret := PrivKey1, public := PubKey1} = libp2p_crypto:generate_keys(ecc_compact),
+    SigFun1 = libp2p_crypto:mk_sig_fun(PrivKey1),
+
+    PeerMap = #{pubkey => libp2p_crypto:pubkey_to_bin(PubKey1),
+                listen_addrs => ["/ip4/8.8.8.8/tcp/1234"],
+                connected => [],
+                nat_type => static,
+                signed_metadata => #{<<"hello">> => <<"world">>}},
+    Peer = libp2p_peer:from_map(PeerMap, SigFun1),
+    ?assertEqual(<<"world">>, libp2p_peer:signed_metadata_get(Peer, <<"hello">>, <<"dlrow">>)),
```
madninja: you're not encoding and decoding the created peer to verify that the proto definition works

src/peerbook/libp2p_peer.proto
```diff
@@ -16,6 +16,7 @@ message peer {
     int64 timestamp = 5;
     map<string, association_list> associations = 6;
     bytes network_id = 7;
+    map<string, bytes> signed_metadata = 8;
```
madninja: I see you're using binaries for the keys below. Is this correct?

Vagabond: I *think* so? I'm really not sure though?

madninja: If you’re using a binary for a key you should change the type in the proto. And add a round trip test

Vagabond: I kind of want a string (in a binary) for the key and a richer way to encode the value (so we can have integer/floats/strings/binaries), I'm jut not good at protobuffs

test/peer_SUITE.erl
Vagabond: Yeah, I will do that.

src/peerbook/libp2p_peer.proto
madninja: Pretty sure a string in a binary is still a binary in protobuf

Vagabond: You can't use `bytes` as a key, so they're strings forever.

test/peer_SUITE.erl
Vagabond: This has been done.

src/peerbook/libp2p_peer.erl
```diff
@@ -368,8 +370,11 @@ decode_unsafe(Bin) ->
 %% throws an error if the peer or one of it's associations can't be
 %% verified
 -spec verify(peer()) -> true.
-verify(Msg=#libp2p_signed_peer_pb{peer=Peer=#libp2p_peer_pb{associations=Assocs}, signature=Signature}) ->
+verify(Msg=#libp2p_signed_peer_pb{peer=Peer0=#libp2p_peer_pb{associations=Assocs, signed_metadata=MD}, signature=Signature}) ->
+    Peer = Peer0#libp2p_peer_pb{signed_metadata=lists:usort(MD)},
+    ct:pal("Verify: Peer ~p", [Peer]),
```
madninja: Remove `ct` bits?

PR #199, opened 2019-08-09
Madninja/session addr info

madninja:
Caches session addr_info information in the swarm's ets table to allow lookup even if the session is not alive

### DIFF
--- a/src/libp2p_config.erl
+++ b/src/libp2p_config.erl
@@ -272,8 +272,8 @@ insert_addr_info(TID, Kind, Pid, AddrInfo) ->
 
 -spec lookup_addr_info(ets:tab(), atom(), pid()) -> {ok, {string(), string()}} | false.
 lookup_addr_info(TID, Kind, Pid) ->
-    case ets:lookup(TID, {{?ADDR_INFO, Kind, '$1'}, Pid}) of
-        [AddrInfo]  -> {ok, AddrInfo};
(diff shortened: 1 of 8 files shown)


src/libp2p_config.erl
```diff
@@ -247,6 +249,35 @@ lookup_session_addrs(TID, Pid) ->
...
+-spec insert_addr_info(ets:tab(), atom(), pid(), {string(), string()}) -> true.
+insert_addr_info(TID, Kind, Pid, AddrInfo) ->
+    %% Insert in the form that remove_pid understands to ensure that
+    %% addr info gets removed for removed pids regardless of what kind
+    %% of addr_info it is
+    ets:insert(TID, {{?ADDR_INFO, Kind, AddrInfo}, Pid}).
+
+-spec lookup_addr_info(ets:tab(), atom(), pid()) -> {ok, {string(), string()}} | false.
+lookup_addr_info(TID, Kind, Pid) ->
+    case ets:lookup(TID, {{?ADDR_INFO, Kind, '$1'}, Pid}) of
```
Vagabond: Should we do an is_process_alive check too?

madninja: We don't do it anywhere else either?

remove_pid causes the addr_info entries to be removed from the table, and that's called by the swarm_server when the session goes away (monitored)

PR #202, opened 2019-08-20
Add randomness to relay peer selection

Vagabond:
When selecting a relay, if everyone has the same relay selection
strategy, and the top relay is misbehaving, the network can become
partitioned. To avoid this, do the normal filtering and sorting of
relays, but start attempting relays at a random index into the list.
This should prevent all nodes selecting the same relay peer or
reconnecting to it over and over again.


Vagabond:
Ignoring the dialyzer issues for now as they're not related to this branch.

PR #203, opened 2019-08-20
Macpie/ch4645/libp2p should have a relay limit

macpie:


evanmcc:
I'm confused, don't we already have some sort of relay limit code in?

PR #205, opened 2019-08-22
Macpie/ch4600/improve docs

macpie:


macpie:
Will do a little doc improvement everyday

PR #209, opened 2019-08-29
add group mgr to de-race group starting

evanmcc:
this will eventually grow to manage more aspects of the groups, but this bit is important enough to add on its own.  also fixes the relcast tests b/c of api breakage.

### DIFF
--- a/src/group/libp2p_group_mgr.erl
+++ b/src/group/libp2p_group_mgr.erl
@@ -1,0 +1,106 @@
+-module(libp2p_group_mgr).
+
+-behaviour(gen_server).
+
+%% API
+-export([
+         start_link/1,
+         mgr/1,
+         add_group/5,
+         remove_group/3,
+         force_gc/1
+        ]).
+
+%% gen_server callbacks
+-export([init/1, handle_call/3, handle_cast/2, handle_info/2,
+         terminate/2, code_change/3]).
+
+-define(SERVER, ?MODULE).
+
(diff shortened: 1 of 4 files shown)


src/group/libp2p_group_mgr.erl
```diff
@@ -0,0 +1,106 @@
...
+-record(state, {}).
+
+%%%===================================================================
+%%% API
+%%%===================================================================
+
+start_link(TID) ->
+    gen_server:start_link(?MODULE, [TID], []).
+
+mgr(TID) ->
```
macpie: Maybe put that and `libp2p_config`. As all the other ets related stuff are in there

src/group/libp2p_group_mgr.erl
```diff
@@ -0,0 +1,106 @@
...
+    ets:lookup_element(TID, ?SERVER, 2).
+
+%% these are a simple wrapper around swarm add group to prevent races
+add_group(Mgr, TID, GroupID, Module, Args) ->
+    gen_server:call(Mgr, {add_group, TID, GroupID, Module, Args}, infinity).
```
madninja: Why not hold on to the TID? You already get it in init

src/group/libp2p_group_mgr.erl
evanmcc: I'm not sure that that's the right place for this, I'm just doing what the group server does.

PR #212, opened 2019-08-30
Don't update peer right away on register/unregister session

madninja:
The working assumption is that we register/unregister session a _lot_ so updating "this" peer loads up the signing subsystem (done over i2c on production hotspots). This change relies on the peer_timeout to refresh the peer record and send it out.


madninja:
Turns out relcast groups were loading up the signing system way more than this. Closing this PR until we get this problem again

PR #215, opened 2019-09-04
accept_stream exception handling

madninja:
gossip_stream was not returning the right response on errors. We just never saw it since the gossip_server is always running. 

ack_stream is enhanced to catch errors since the relcast group may already be gone when a stream is inbound

### DIFF
--- a/src/group/libp2p_ack_stream.erl
+++ b/src/group/libp2p_ack_stream.erl
@@ -42,13 +42,23 @@ server(Connection, Path, _TID, Args) ->
     libp2p_framed_stream:server(?MODULE, Connection, [Path | Args]).
 
 init(server, Connection, [Path, AckModule, AckState | _]) ->
-    case AckModule:accept_stream(AckState, self(), Path) of
+    %% Catch errors from the handler module since the handling group
+    %% may have already stopped or crashed.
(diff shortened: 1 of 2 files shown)


src/group/libp2p_ack_stream.erl
```diff
@@ -42,13 +42,23 @@ server(Connection, Path, _TID, Args) ->
...
+    case (catch AckModule:accept_stream(AckState, self(), Path)) of
         {ok, AckRef} ->
             libp2p_connection:set_idle_timeout(Connection, ?ACK_STREAM_TIMEOUT),
             {ok, #state{connection=Connection,
                         ack_ref=AckRef, ack_module=AckModule, ack_state=AckState}};
         {error, Reason} ->
-            {stop, {error, Reason}}
+            {stop, {error, Reason}};
+        {'EXIT', {noproc, _}} ->
+            %% Handle noproc silently since it's pretty common for a
```
evanmcc: shouldn't the workers be linked to or monitoring their group server?

madninja: The workers and their streams are.. those are clients though.. This is a server clause. The stream handlers are registered at the swarm level. 

So either someone isn’t deregistering the stream handlers for a given path, or there’s a slew of them coming in before the de-registration is complete.

PR #216, opened 2019-09-05
Log yamux session failure as notice

vihu:


src/libp2p_yamux_session.erl
```diff
@@ -155,7 +155,9 @@ handle_info({inert_read, _, _}, State=#state{connection=Connection}) ->
             lager:notice("session closed"),
             {stop, normal, State};
         {error, Reason} ->
-            lager:error("Session header read failed: ~p ", [Reason]),
+            %% Apparently this happens a LOT
+            %% just flag a notice, don't pollute the logs with an error
+            lager:notice("Session header read failed: ~p ", [Reason]),
```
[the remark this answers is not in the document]
vihu: I was contemplating going down that path. Will talk to @madninja and filter unhelpful logs.

vihu:
Just removed any logging for `enotconn` since that's logged like 90% of the times anyway.

src/libp2p_yamux_session.erl
```diff
@@ -155,7 +155,15 @@ handle_info({inert_read, _, _}, State=#state{connection=Connection}) ->
             lager:notice("session closed"),
             {stop, normal, State};
         {error, Reason} ->
-            lager:error("Session header read failed: ~p ", [Reason]),
+            %% Apparently this happens a LOT
+            %% just flag a notice, don't pollute the logs with an error
+            case Reason == enotconn of
```
madninja: just do `{error, enotconn} -> ` and a next clause with `{error, reason}`. No need to nest case and have a boolean case statement  like this

vihu: Addresses: a07238d

PR #221, opened 2019-09-16
libp2p_crypto factored into it's own repo

madninja:


rebar.config
```diff
@@ -23,6 +23,7 @@
     {ecc_compact, "1.0.4"},
     {enacl, "0.17.2"},
     {erl_base58, "0.0.1"},
+    {libp2p_crypto, "1.0.1"},
```
macpie: pushed to hex! nice!

PR #222, opened 2019-09-25
Macpie/ch4964/peerbook corruption prevents any new keys

macpie:


vihu:
why are the tests mad though?

vihu:
peerbook_SUITE worked locally for me though

PR #224, opened 2019-10-03
add old group GC to the group manager

evanmcc:
it now owns old group deletion, and relcast groups just stop and start without worrying about storage

### DIFF
--- a/src/group/libp2p_group_mgr.erl
+++ b/src/group/libp2p_group_mgr.erl
@@ -134,10 +134,11 @@ gc(Predicate, Dir) ->
     case file:list_dir(Dir) of
         {ok, Groups} ->
             %% filter using predicate
-            Dels = lists:filter(Predicate, Groups),
+            Dels = lists:filter(Predicate, lists:sublist(Groups, 100)),
             lager:debug("groups ~p dels ~p", [Groups, Dels]),
             %% delete.
-            lists:foreach(fun(Grp) -> rm_rf(Dir ++ "/" ++ Grp) end, Dels);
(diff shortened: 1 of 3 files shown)


src/group/libp2p_group_mgr.erl
```diff
@@ -86,15 +93,25 @@ handle_call({remove_group, GroupID}, _From, #state{tid = TID} = State) ->
...
+handle_call(force_gc, _From, #state{group_deletion_predicate = Predicate,
+                                    storage_dir = Dir} = State) ->
+    lager:info("forcing gc"),
+    Reply = gc(Predicate, Dir),
+    {reply, Reply, State};
 handle_call(_Request, _From, State) ->
     lager:warning("unexpected call ~p from ~p", [_Request, _From]),
-    Reply = ok,
-    {reply, Reply, State}.
+    {noreply, State}.
```
Vagabond: Won't this deadlock the caller (or trigger a timeout)?

src/group/libp2p_group_mgr.erl
```diff
@@ -19,15 +19,17 @@
 
 -record(state,
         {
-         tid :: term()
+         tid :: term(),
+         group_deletion_predicate :: function(),
```
Vagabond: Can we get a better spec here with the input types and the return type?

src/group/libp2p_group_mgr.erl
evanmcc: yeah.  I do this to call out bad calls by forcing them to error, hopefully in test.

PR #228, opened 2019-10-18
update relcast for new rocks

evanmcc:


Vagabond:
1.4.0 is out now, do we still want to upgrade rocks (we have 2 patches I can rebase on top of the 1.4.0 tag if we want to upgrade).

PR #231, opened 2019-11-30
Adt/async signed metadata

Vagabond:


src/peerbook/libp2p_peerbook.erl
```diff
@@ -359,41 +368,33 @@ mk_this_peer(CurrentPeer, State=#state{tid=TID}) ->
...
-                                        fun() -> #{} end),
-    %% if the metadata fun crashes, simply return an empty map
-    MetaData = try MetaDataFun() of
-                   Result ->
-                       Result
-               catch
-                   _:_ -> #{}
-               end,
-    libp2p_peer:from_map(#{ pubkey => SwarmAddr,
+    {MetaData, State1} = get_signed_metadata(State),
```
madninja: I don't quite get why you need to get metadata back from the get_signed_metadata. It's implementation _always_ returns the metadata in `#state{}`.. and that metadata is refreshed when the spawned pid returns. Why not just always pick up the metadata from state?

src/peerbook/libp2p_peerbook.erl
```diff
@@ -562,3 +563,27 @@ install_gossip_handler(TID, Handle) ->
...
+get_signed_metadata(State = #state{tid=TID, metadata_ref=MR}) ->
+    case MR of
+        undefined ->
+            MetaDataFun = libp2p_config:get_opt(libp2p_swarm:opts(TID), [libp2p_peerbook, signed_metadata_fun],
+                                                fun() -> #{} end),
+
+            Parent = self(),
+            {_, Ref} = spawn_monitor(fun() ->
+                                             %% if the metadata fun crashes, use the old metadata
+                                             try MetaDataFun() of
```
madninja: So maybe remove the try catch and let spawn_monitor do it's job to leave the old metadata alone?

src/peerbook/libp2p_peerbook.erl
```diff
@@ -562,3 +563,27 @@ install_gossip_handler(TID, Handle) ->
             libp2p_group_gossip:add_handler(G,  ?GOSSIP_GROUP_KEY, {?MODULE, Handle}),
             G
     end.
+
+get_signed_metadata(State = #state{tid=TID, metadata_ref=MR}) ->
+    case MR of
+        undefined ->
+            MetaDataFun = libp2p_config:get_opt(libp2p_swarm:opts(TID), [libp2p_peerbook, signed_metadata_fun],
```
madninja: Do we have to look up metadata_fun every time? Why not just get it on peer book init and pass store it in state?

src/peerbook/libp2p_peerbook.erl
madninja: or store the spawned pid in the state if the spawn is not going to crash because you try catch everything and send an error back to the parent so it can clear the stored pid

madninja: I.e. ```-spec get_signed_metadata(#state{}) -> #state{}.
get_signed_metadata(State=#state{metadata_fun=MetaDataFun}) ->
    case State#state.metadata_ref of
        undefined ->
            Parent = self(),
            {_, Ref} = spawn_monitor(fun() ->
                                             Parent ! {signed_metadata, MetaDataFun()}
                                     end),
            State#state{metadata_ref = Ref};
        _ ->
            State
    end.
```

File: src/peerbook/libp2p_peerbook.erl (hunk 3 above)
Comment (Vagabond) [reply src/peerbook/libp2p_peerbook.erl]: it;s an ETS lookup, it should be fast?

File: src/peerbook/libp2p_peerbook.erl (hunk 1 above)
Comment (Vagabond) [reply src/peerbook/libp2p_peerbook.erl]: Yeah, not sure why I did it this way, will rework.

File: src/peerbook/libp2p_peerbook.erl (hunk 2 above)
Comment (Vagabond) [reply src/peerbook/libp2p_peerbook.erl]: Right, this is a holdover from the old code.

File: src/peerbook/libp2p_peerbook.erl (hunk 3 above)
Comment (madninja) [reply src/peerbook/libp2p_peerbook.erl]: but there's no need to?

File: src/peerbook/libp2p_peerbook.erl [hunk 4]
```diff
@@ -562,3 +564,21 @@ install_gossip_handler(TID, Handle) ->
             libp2p_group_gossip:add_handler(G,  ?GOSSIP_GROUP_KEY, {?MODULE, Handle}),
             G
     end.
+
+get_async_signed_metadata(State = #state{tid=TID, metadata_ref=MR}) ->
```
madninja: Maybe just make MR==undefined a function head?

PR #236, opened 2019-12-20
Don't put RFC1918 addresses in peerbook entries by default

Vagabond:
Additionally, don't gossip peer entries if they do not have a listen
address or any peer connections.

Finally add some commented out code to reject peers that send us
peerbook updates containing RFC1918 addresses if we're configured not to
use them. This should be activated later once the network has fully
upgraded to not publishing them.

### DIFF
--- a/src/peerbook/libp2p_peerbook.erl
+++ b/src/peerbook/libp2p_peerbook.erl
@@ -61,6 +61,12 @@
 %% minutes (in milliseconds)
 -define(DEFAULT_GOSSIP_PEERS_TIMEOUT, 30 * 60 * 1000).
 
+-ifdef(TEST).
+-define(DEFAULT_PEERBOOK_ALLOW_RFC1918, true).
+-else.
+-define(DEFAULT_PEERBOOK_ALLOW_RFC1918, false).
+-endif.
+
 %%
 %% API
 %%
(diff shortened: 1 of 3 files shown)


src/peerbook/libp2p_peerbook.erl
```diff
@@ -450,6 +476,33 @@ notify_new_peers(NewPeers, State=#state{notify_timer=NotifyTimer, notify_time=No
...
+    ListenAddrs = case AllowRFC1918 of
+                      true ->
+                          %% everything is ok
+                          ListenAddrs0;
+                      false ->
+                          filter_rfc1918_addresses(ListenAddrs0)
+                  end,
+    ListenAddrs /= [].
+
+filter_rfc1918_addresses(ListenAddrs) ->
```
madninja: Maybe move part of this to transport_tcp (taking a p2p adds) so that you can use the transport_tcp function as filter instead of doing it here? Basically have another head to rfc1918

src/peerbook/libp2p_peerbook.erl
```diff
@@ -365,7 +376,16 @@ terminate(_Reason, _State) ->
 -spec mk_this_peer(libp2p_peer:peer() | undefined, #state{}) -> {ok, libp2p_peer:peer()} | {error, term()}.
 mk_this_peer(CurrentPeer, State=#state{tid=TID}) ->
     SwarmAddr = libp2p_swarm:pubkey_bin(TID),
-    ListenAddrs = libp2p_config:listen_addrs(TID),
+    Opts = libp2p_swarm:opts(TID),
+    AllowRFC1918 = libp2p_config:get_opt(Opts, [?MODULE, allow_rfc1918], ?DEFAULT_PEERBOOK_ALLOW_RFC1918),
```
madninja: Maybe move allow to a function and use it her and in the put/3 function usage above? Because the way it's coded now we allow rfc1918 in put but don't mk_peer one?

Vagabond: This is intentional because we need to transition through the old nodes publishing 1918 addresses, once we've upgraded the network we can enable rejection of peers with 1918 addresses in them.

madninja: AH yes, that makes total sense.. Would still be nice not to repeat ourselves though

PR #242, opened 2020-01-14
Remove the gossip server's peer cache

Vagabond:
The intent of the peer cache was to quickly provide a random peer for a
gossip worker by storing all the peerbook addresses as p2p addresses in
the gossip server's state. This doesn't scale at all AND storing the p2p
address instead of the pubkey bin means that we pay the 4x memory
overhead of storing strings (32 bit code points) and can't benefit from
binary sharing.

This patch adds peerbook:random/1 and random/2 for quickly (2000
microseconds) getting a random peer from the peerbook. With random/2,
a list of peer addresses to exclude from selection is provided.


madninja:
💯

PR #243, opened 2020-01-14
Cap the number of relay candidates stored in relay server's state

Vagabond:


src/relay/libp2p_relay_server.erl
```diff
@@ -256,7 +257,7 @@ sort_peers(Peers0, SwarmPubKeyBin, State) ->
         libp2p_peer:has_public_ip(Peer) andalso
         not lists:member(libp2p_peer:pubkey_bin(Peer), State#state.banlist)
     end, Peers0),
-    lists:sort(fun sort_peers_fun/2, shuffle(Peers1)).
+    lists:sublist(lists:sort(fun sort_peers_fun/2, shuffle(Peers1)), ?MAX_PEERS).
```
evanmcc: you should shuffle then sublist then sort, imo

Vagabond: we want to keep the top candidates, though?

PR #254, opened 2020-02-07
store similar peers, add debugging to relcast exits.

evanmcc:
Currently, although we've gotten a new peer, we gate both storage and gossip on the same filter.  This PR stores peers as long as they're an updated, but does not gossip them, in the grounds that we've already paid the price of deserializing and testing the peer, so we might as well take the updated information.

### DIFF
--- a/src/peerbook/libp2p_peerbook.erl
+++ b/src/peerbook/libp2p_peerbook.erl
@@ -77,25 +77,40 @@ put(#peerbook{tid=TID, stale_time=StaleTime}=Handle, PeerList) ->
     ThisPeerId = libp2p_swarm:pubkey_bin(TID),
     %% XXX uncomment this to reject any peers publishing RFC1918 addresses once the network has transitioned over
     AllowRFC1918 = true, %% is_rfc1918_allowed(TID),
-    NewPeers = lists:filter(fun(NewPeer) ->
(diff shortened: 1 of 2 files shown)


src/peerbook/libp2p_peerbook.erl
```diff
@@ -77,22 +77,40 @@ put(#peerbook{tid=TID, stale_time=StaleTime}=Handle, PeerList) ->
...
+                             {ok, ExistingPeer} ->
+                                 %% Only store peers that are not _this_ peer,
+                                 %% are newer than what we have,
+                                 %% are not stale themselves
+                                 case NewPeerId /= ThisPeerId
+                                     andalso (AllowRFC1918 orelse not libp2p_peer:has_private_ip(NewPeer))
+                                     andalso libp2p_peer:supersedes(NewPeer, ExistingPeer)
+                                     andalso not libp2p_peer:is_stale(NewPeer, StaleTime)
+                                     andalso libp2p_peer:network_id_allowable(NewPeer, libp2p_swarm:network_id(TID))
+                                     andalso not libp2p_peer:is_similar(NewPeer, ExistingPeer) of
```
madninja: Why not just remove the actual is_similar negative check ? If this is needed we need to refactor this a bit

Vagabond: Yeah, I think this was left in as an oversight, will remove it.

PR #258, opened 2020-02-13
don't gossip seed node peer

evanmcc:


src/peerbook/libp2p_peerbook.erl
```diff
@@ -502,22 +502,27 @@ mk_this_peer(CurrentPeer, State=#state{tid=TID}) ->
...
-            case mk_this_peer(OldPeer, State) of
-                {ok, NewPeer} ->
-                    case libp2p_peer:is_similar(NewPeer, OldPeer) of
-                        true -> State;
-                        false -> update_this_peer({ok, NewPeer}, get_async_signed_metadata(State))
-                    end;
-                {error, Error} ->
-                    lager:notice("Failed to make peer: ~p", [Error]),
-                    State
-            end
```
madninja: Miner env variables in here seems like some serious name leakage? Can we change that?

evanmcc: yeah, this was just something quick, I'll change it.

PR #259, opened 2020-02-17
catch errors when starting relcast, set cannot start as status

andymck:
If start_relcast fails for any reason it may have unexpected knock on effects higher up.  With this fix we catch any error and set the relcast servers status too 'cannot start'.

This addresses the scenario whereby a consensus miner restarts ( for reason unknown ).  During startup it attempts to load the ondisk consensus group data which for some reason may now be missing, contain bad data or be malformed.  When this happens the relcast hbbft_data reverts to that provided to the miner_hbbft_handler and in this case the security key here will be undefined ( see timeout msg handling in miner_hbbft_handler ) .  This leads to a function clause crash whilst starting the relcast server as tpke_privkey cannot serialize the undefined key.  

The function clause error gets set as the relcast server's #state.store value.

The consensus mgr after starting the group will call 'wait_for_group' which in turn calls the relcast server to get the status which is that servers #state.store value.  At this point the consensus mgr is returned the function clause error as the status value and as the mgr does not obviously handle this response the consensus mgr crashes out.  The miner sup tears ev thing else down and restarts and we then enter the restart loop.

The fix here is to catch any error whilst starting relcast, upon doing so we set the #state.store field to 'cannot_start' and as this is handled by the consensus_mgr's wait_for_group call the crash is prevented.  

We could have fixed on the consensus mgr end and handled the error status but its better to not let errors like this propagate upstream by becoming state var values.

### DIFF
--- a/src/group/libp2p_group_relcast_server.erl
+++ b/src/group/libp2p_group_relcast_server.erl
@@ -337,7 +337,7 @@ handle_info({start_workers, Targets}, State=#state{group_id=GroupID, group_keys=
                                                                 {keys, GroupKeys}]}),
     {noreply, State#state{workers=start_workers(Targets, State)}};
 handle_info({start_relcast, Handler, HandlerArgs, RelcastArgs, SelfIndex, Addrs}, State) ->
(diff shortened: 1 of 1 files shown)


Vagabond:
This looks pretty reasonable

PR #264, opened 2020-03-17
don't die to exceptions when signing peer

evanmcc:
if we get a transient ecc failure here, we can kill the peerbook and cause all sorts of restart problems (currently the blockchain_swarm can't be safely restarted).  So we handle the catch here and turn it into a normal error so we can retry later.

### DIFF
--- a/src/peerbook/libp2p_peer.erl
+++ b/src/peerbook/libp2p_peer.erl
@@ -415,11 +415,15 @@ verify(Msg=#libp2p_signed_peer_pb{peer=Peer0=#libp2p_peer_pb{associations=Assocs
 sign_peer(Peer0 = #libp2p_peer_pb{signed_metadata=MD}, SigFun) ->
     Peer = Peer0#libp2p_peer_pb{signed_metadata=lists:usort(MD)},
     EncodedPeer = libp2p_peer_pb:encode_msg(Peer),
-    case SigFun(EncodedPeer) of
+    try SigFun(EncodedPeer) of
         {error, Error} ->
             {error, Error};
         Signature ->
             {ok, #libp2p_signed_peer_pb{peer=Peer, signature=Signature}}
+    catch C:E ->
+            %% probably a timeout
+            lager:info("signing peer failed: ~p:~p", [C, E]),
+            {error, sign_failure}
     end.
 
 encode_map(Map) ->


src/libp2p_swarm_sup.erl
```diff
@@ -87,7 +87,7 @@ init([Name, Opts]) ->
     ets:insert(TID, {?ADDRESS, libp2p_crypto:pubkey_to_bin(PubKey)}),
     GroupDeletePred = libp2p_config:get_opt(Opts, group_delete_predicate, fun(_) -> false end),
 
-    SupFlags = {one_for_all, 3, 10},
+    SupFlags = {one_for_one, 3, 10},
```
evanmcc: @madninja @macpie can you think of any reason why we shouldn't make this change?  peerbook failures shouldn't really take the rest of the swarm components down, from what I can tell, but I don't know if this was done for a particular reason.

Vagabond: The peerbook will be it's own thing shortly once @andymck is done with his refactor, too.

madninja: My main concern would be that the children of this sup aren't ready to be restarted without the whole supervisor coming back up to give them all the handles/state they need.

andymck: yeah i am changing this area.  the peerbook is in its own tree, and the swarm components in their own tree.  Both trees reside under a top level swarm specific 'super sup'.  Still a few issues to iron out atm with handles but I am currently using named tables ( so the ETS TID is now just the swarm name ) and thus any process can get the peerbook handle as long as they know the swarm name...

andymck: I am also considering using registered names for core processes,  based around the swarm name,  to avoid having to know the PID ( for example for the peerbook )

trevorprater:
Is this related to the `command_too_small` error that arises from the ecc interaction?

trevorprater:
e.g..

```
{{badmatch,{error,{ecc_command_too_small,0}}},[{miner_keys,'-keys/1-fun-1-',1,[{file,"miner_keys.erl"},{line,111}]},{blockchain_state_channel_offer_v1,sign,2,[{file,"blockchain_state_channel_offer_v1.erl"},{line,82}]},{blockchain_state_channels_client,send_offer,5,[{file,"blockchain_state_channels_client.erl"},{line,629}]},{blockchain_state_channels_client,'-maybe_send_packets/3-fun-5-',5,[{file,"blockchain_state_channels_client.erl"},{line,707}]},{lists,foldl,3,[{file,"lists.erl"},{line,1263}]},{blockchain_state_channels_client,maybe_send_packets,3,[{file,"blockchain_state_channels_client.erl"},{line,705}]},{blockchain_state_channels_client,handle_cast,2,[{file,"blockchain_state_channels_client.erl"},{line,152}]},{gen_server,try_dispatch,4,[{file,"gen_server.erl"},{line,637}]}]}```

**Comment by trevorprater** [comment]
... and

```
2021-03-20 08:36:33.905 1 [info] <0.1199.0>@libp2p_peer:sign_peer:436 signing peer failed: error:{badmatch,{error,{ecc_command_too_small,0}}}
2021-03-20 08:36:33.905 1 [notice] <0.1199.0>@libp2p_peerbook:update_this_peer:565 Failed to make peer: sign_failure
```

madninja:
> e.g..
> 
> ```
> {{badmatch,{error,{ecc_command_too_small,0}}},[{miner_keys,'-keys/1-fun-1-',1,[{file,"miner_keys.erl"},{line,111}]},
> {blockchain_state_channel_offer_v1,sign,2,[{file,"blockchain_state_channel_offer_v1.erl"},{line,82}]},
> {blockchain_state_channels_client,send_offer,5,[{file,"blockchain_state_channels_client.erl"},{line,629}]},
> {blockchain_state_channels_client,'-maybe_send_packets/3-fun-5-',5,[{file,"blockchain_state_channels_client.erl"},{line,707}]},
> {lists,foldl,3,[{file,"lists.erl"},{line,1263}]},{blockchain_state_channels_client,maybe_send_packets,3,
> [{file,"blockchain_state_channels_client.erl"},{line,705}]},{blockchain_state_channels_client,handle_cast,2,
> [{file,"blockchain_state_channels_client.erl"},{line,152}]},{gen_server,try_dispatch,4,[{file,"gen_server.erl"},{line,637}]}]}
> ```

No it is not.. You seem to have issues communicating with the ECC chip which should indeed be fatal

PR #266, opened 2020-04-03
don't crash the nat server on bad cache insert

evanmcc:
having some issues where we can't insert to the cache for whatever reason and the nat ever restarts until the node restarts.  this gives the cache write more time to complete, and makes it so that the nat server doesn't care about the outcome.

### DIFF
--- a/src/libp2p_cache.erl
+++ b/src/libp2p_cache.erl
@@ -51,7 +51,7 @@ reg_name(TID)->
 %%--------------------------------------------------------------------
 -spec insert(pid(), any(), any()) -> ok | {error, any()}.
 insert(Pid, Key, Value) ->
-    gen_server:call(Pid, {insert, Key, Value}).
+    gen_server:call(Pid, {insert, Key, Value}, 30000).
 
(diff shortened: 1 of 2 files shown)


src/libp2p_cache.erl
```diff
@@ -51,7 +51,7 @@ reg_name(TID)->
 %%--------------------------------------------------------------------
 -spec insert(pid(), any(), any()) -> ok | {error, any()}.
 insert(Pid, Key, Value) ->
-    gen_server:call(Pid, {insert, Key, Value}).
+    gen_server:call(Pid, {insert, Key, Value}, 30000).
```
andymck, line 54: do inserts and lookups need to be serialised here ?  If not then the sync calls could be eliminated and just have a simple function runs in the caller context.  Obviously this is dependant upon the caller knowing the table name but as this is just the swarm name that wont be too difficult.  With my open libp2p peerbook changes, TID is in fact just the swarm name which is passed to all processes currently....so this would be an easy change on that branch if we dont need to serialise things....

PR #269, opened 2020-04-23
up the staleness time for relays and add stale filtering to random

evanmcc:


Vagabond:
Tests all pass locally here

PR #270, opened 2020-04-23
Track group workers by ref, not pid, so restarts can be handled

Vagabond:
The current gossip server was only tracking workers by pid, which means
if they crashed the supervisor would restart them, but the gossip server
would not notice they'd been restarted. This would prevent those gossip
connections from being sent outbound messages and degrade the gossip
group. This PR instead tracks the gossip workers by their Ref, which is
is in the supervision spec, so the restarted process will have the same
Ref.

### DIFF
--- a/src/group/libp2p_group_gossip_server.erl
+++ b/src/group/libp2p_group_gossip_server.erl
@@ -134,9 +134,9 @@ handle_cast({handle_data, StreamPid, Key, ListOrData}, State=#state{}) ->
 
 handle_cast({add_handler, Key, Handler}, State=#state{handlers=Handlers}) ->
     {noreply, State#state{handlers=maps:put(Key, Handler, Handlers)}};
-handle_cast({request_target, inbound, WorkerPid}, State=#state{}) ->
+handle_cast({request_target, inbound, WorkerPid, _Ref}, State=#state{}) ->
     {noreply, stop_inbound_worker(WorkerPid, State)};
-handle_cast({request_target, peerbook, WorkerPid}, State=#state{tid=TID}) ->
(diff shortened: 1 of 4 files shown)


src/group/libp2p_group_gossip_server.erl
```diff
@@ -298,27 +298,39 @@ connections(Kind, #state{workers=Workers}) ->
...
+            %% the ref is stable across restarts, so use that as the lookup key
+            case lookup_worker(WorkerRef, #worker.ref, State) of
+                Worker=#worker{kind=seed, target=SelectedAddr, pid=P} when SelectedAddr /= undefined ->
                     %% don't give up on the seed nodes in case we're entirely offline
                     %% we need at least one connection to bootstrap the swarm
                     ClientSpec = {?GROUP_PATH, {libp2p_gossip_stream, [?MODULE, self()]}},
-                    libp2p_group_worker:assign_target(WorkerPid, {SelectedAddr, ClientSpec});
+                    libp2p_group_worker:assign_target(WorkerPid, {SelectedAddr, ClientSpec}),
+                    %% check if this worker got restarted
+                    case WorkerPid /= P of
```
madninja, line 312: mind calling `P` StoredWorkerPid or something to make it clearer?

Vagabond: Fixed

andymck:
@Vagabond really good find.  I have been looking at this code ( unfortunately didnt see this ) but had been thinking this code is somewhat fragile, the workers would be better managed via a dedicated pool or process registry.  Something to consider depending which way we go with libp2p

PR #273, opened 2020-04-28
dialyzer surgery

evanmcc:
we have to abuse the type system a little here to get it to ignore the horrors we're committing for the pre-decode speedup.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -65,11 +65,6 @@
 
 {shell, [{apps, [lager, ranch]}]}.
 
-{dialyzer, [
-    {warnings, [unknown]},
-    {plt_apps,all_deps}
-]}.
-
 {xref_checks,[
     undefined_function_calls,
     undefined_functions, %locals_not_used,
--- a/src/peerbook/libp2p_peerbook.erl
+++ b/src/peerbook/libp2p_peerbook.erl
@@ -333,9 +333,10 @@ lookup_association(Handle=#peerbook{}, AssocType, AssocAddress) ->
 %% Gossip Group
 %%
 
--spec handle_gossip_data(pid(), binary(), peerbook()) -> noreply.
+
+%% the any here is because we had to do bad things to the typesystem
+-spec handle_gossip_data(pid(), binary() | [libp2p_peer:peer()], peerbook()) -> noreply.
 handle_gossip_data(_StreamPid, DecodedList, Handle) ->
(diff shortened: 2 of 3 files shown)


src/peerbook/libp2p_peerbook.erl
```diff
@@ -333,9 +332,10 @@ lookup_association(Handle=#peerbook{}, AssocType, AssocAddress) ->
 %% Gossip Group
 %%
 
--spec handle_gossip_data(pid(), binary(), peerbook()) -> noreply.
+
+%% the any here is because we had to do bad things to the typesystem
```
madninja, line 336: I was looking for an any but it's not?

rebar.config
```diff
@@ -65,11 +65,6 @@
 
 {shell, [{apps, [lager, ranch]}]}.
 
-{dialyzer, [
-    {warnings, [unknown]},
```
madninja, line 69: I wasn't expecting you'd have to strip this section out after you fixed the abuse of handle_gossip_data?

src/peerbook/libp2p_peerbook.erl
```diff
@@ -88,8 +88,7 @@ put(#peerbook{tid=TID, stale_time=StaleTime}=Handle, PeerList0, Prevalidated) ->
             false -> lists:filter(fun libp2p_peer:verify/1, PeerList0)
         end,
     ThisPeerId = libp2p_swarm:pubkey_bin(TID),
-    %% XXX uncomment this to reject any peers publishing RFC1918 addresses once the network has transitioned over
-    AllowRFC1918 = true, %% is_rfc1918_allowed(TID),
+    AllowRFC1918 = is_rfc1918_allowed(TID),
```
Vagabond, line 91: We should check we don't have many peers advertising RFC1918 addresses before we merge this.

src/peerbook/libp2p_peerbook.erl
Vagabond, line 336: It's not an any?

rebar.config
evanmcc: I had to get rid of this to get rid of the throttle errors I was seeing.  maybe I didn't need to?

src/peerbook/libp2p_peerbook.erl
evanmcc: I'll check that tomorrow morning.

PR #275, opened 2020-05-05
Use named_table for swarm ETS

andymck:
By using named_table for the swarm ETS, the TID will always be the same as that of the swarm name.  This provides the benefit that any process or component with access to the swarm name can access the ETS table without the need to perform and sync calls to a process holding the ETS tid


andymck:
bah cut from wrong feature branch and not master

PR #276, opened 2020-05-05
use named table for swarm ets

andymck:
By using named_table for the swarm ETS, the TID will always be the same as that of the swarm name. This provides the benefit that any process or component with access to the swarm name can access the ETS table without the need to perform sync calls to a process holding the ETS TID

Associated blockchain core PR: https://github.com/helium/blockchain-core/pull/442
Associated miner PR: https://github.com/helium/miner/pull/384

### DIFF
--- a/src/libp2p_swarm.erl
+++ b/src/libp2p_swarm.erl
@@ -286,7 +286,7 @@ add_connection_handler(TID, Key, {ServerMF, ClientMF}) ->
     libp2p_config:insert_connection_handler(TID, {Key, ServerMF, ClientMF}),
     ok.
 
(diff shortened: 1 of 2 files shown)


evanmcc:
it might be nice to check that there's no pre-existing table with the same name?  it seems bad to double-start under this model.

Vagabond:
> it might be nice to check that there's no pre-existing table with the same name? it seems bad to double-start under this model.

But if the supervisor is the owner, and the supervisor restarts, the ETS table will die?

andymck:
yeah the table will be lost if the sup restarts, so double starting wont be an issue

PR #277, opened 2020-05-08
Accept new stream over existing stream

andymck:

PR #278, opened 2020-05-12
prevent peers both dialing each other 

andymck:
This PR addresses two related problems:

**Problem Description 1**
----------
There is a race condition which can trigger after a _libp2p_swarm:connect/2_ is called.  As part of the connect call the identities of both peers are swapped and peerbook entries are added for the other on each peer.  

It is after the peerbook entries are added where the race can occur.  Each peer's group workers will request a target from the group server.  If they happen to do this at the same time  and be assigned each others as the target ( which in the tests can happen frequently ) the workers will dial a stream to each other.

This will result in two sets of streams being created on and for each of the two peers.  On each peer, when the second stream comes up it will be assigned to the same group worker as the first.  And as there is an existing stream, the _group_worker:handle_assign_stream/2_ function will flip a coin to determine which stream to keep.  As this happens on both peers, should both peers end up with opposite sides of the coin, it will result in both the first and the second streams being closed.  The peers are then left without any active streams to each other.

The reconnect logic will then kick in but in many cases we end up repeating the exact same scenario again after which the group worker will give up connecting/dialing and revert back to targeting.

**Observable impact on tests**

The exact extent of the observable behaviour is determined by how closely together the two sets of streams come up ( there could be network latency on one or both routes or  a swarm running slowly ).  If both peers dials result in streams being assigned in quick succession, the impact is mostly that the tests take longer to advance as they are held up at the _await_gossip_group_ and _await_gossip_streams_ handlers.  As the group workers have a back off and will ultimately revert to targeting after they exceed the connect retry limit, the _await*_ functions can timeout leading to test fails.

If one of the two dials take longer for the streams to be generated the impact on the tests is that they advance quickly past the _await*_ functions and start to exchange msgs via gossip using the first stream.  During this exchange the second stream can come up ( resulting in both streams being closed due to the peers getting opposite sides of the coin toss ) and the tests fail due to gossip msgs being lost.

**Problem Description 2**
----------
In the above scenario after both peers close the streams, the group workers will attempt to reconnect.  If the reconnect attempts result in the same scenario, ie both peers dialing each other, producing two streams, coin toss closing both streams...then the group workers will give up connecting/dialing and revert back to targeting.

At this point targeting will request a new target from the group sever.  However the previously assigned target peer will never be reassigned as the group server has not been told the original assign_target failed and as far as its concerned the target is assigned to a worker ( the group server marks a target as assigned to a worker as part of its assign_target call and does so before the group worker receives the target ).

When this happens the previously assigned peer will never be reassigned to a new group worker.

The observable impact of this via the tests is that the tests never advance beyond the _await_gossip_group_ or _await_gossip_streams_ functions.

Fix for problem 1
----------------
Prevent peers dialing streams to each other.  Assuming both peers are interconnected, only one peer will attempt to dial a stream.  The decision is based on the lexicographically order of both peer's p2p address.  Using the pubkeybin is also possible here.  I cannot use the identity for this decision as we have to account for the scenario whereby peer  A is gossipped new peer C from connected peer B.  Peer A is not connected with peer C and thus not in possession of its identity ( unless I am missing something such as being able to retrieve the identify from the peerbook ?...)

Fix for problem 2
----------------
The group worker, should the connect to the remote peer fail and it reverts to targeting, will now inform the group server the connect attempt failed and it will clear the target for that worker.  This will make the peer available again by the group server for targeting and will be assigned group workers upon subsequent request_target requests.

### DIFF
--- a/src/group/libp2p_group_worker.erl
+++ b/src/group/libp2p_group_worker.erl
@@ -28,6 +28,7 @@
 
 -define(MIN_TARGETING_RETRY_TIMEOUT, 1000).
 -define(MAX_TARGETING_RETRY_TIMEOUT, 10000).
+-define(CONNECT_TIMEOUT, 5000).
 
 -type target() :: {MAddr::string(), Spec::stream_client_spec()}.
 
@@ -231,11 +232,32 @@ connecting(info, connect_retry_timeout, Data=#data{tid=TID,
         false ->
             Parent = self(),
             Pid = erlang:spawn_link(fun() ->
-                case libp2p_swarm:dial_framed_stream(TID, MAddr, Path, M, A) of
-                    {error, Error} ->
-                        Parent ! {connect_error, Error};
-                    {ok, StreamPid} ->
-                        Parent ! {assign_stream, StreamPid}
+                LocalAddr = libp2p_swarm:p2p_address(TID),
+                %% its possible that after exchanging identities two peers will be assigned each other as a target
+                %% and end up dialing each other at pretty much the same time
+                %% resulting in streams being assigned for the first dial, followed by a second set of streams for the second dial
+                %% there is/was then a coin toss ( see handle_assign_stream ) as to which streams win ( we dont want two streams between the same set of peers )
+                %% both peers could get opposite sides of the coin and end up with both streams streams being closed by opposite peers
+                %% to avoid this we will only have one side perform the dialing based on a decision around data available on both ends
+                %% as such, we lexicographically compare the local peer addr with that of the remote peers addr
+                %% peer with highest order gets to do the stream dial
+                case LocalAddr > MAddr of
+                    true ->
+                        lager:debug("(~p) peer ~p will dial a stream to remote peer ~p", [Parent, LocalAddr, MAddr]),
+                        case libp2p_swarm:dial_framed_stream(TID, MAddr, Path, M, A) of
+                            {error, Error} ->
+                                Parent ! {connect_error, Error};
+                            {ok, StreamPid} ->
+                                Parent ! {assign_stream, StreamPid}
+                        end;
+                false ->
+                        %% if our local peer is not performing the dial, we will attempt to connect to the remote peer
+                        %% this is to handle scenarios where we were hoping to dial a peer to which we had not yet connected
+                        %% such as when we are gossiped a new peer from an already connected peer
+                        %% the dial above would have first performed this connection and then proceeded to dial a stream
+                        %% If we dont perform the connect here then we will end up never connecting to gossiped peers with an address with a lower lex order
+                        lager:debug("(~p) peer ~p will not dial a stream to remote peer ~p, but will establish a connection to it if not already done...",[Parent, LocalAddr, MAddr]),
+                        libp2p_transport:connect_to(MAddr, [], ?CONNECT_TIMEOUT, TID)
                 end
             end),
             {keep_state, stop_connect_retry_timer(Data#data{connect_pid=Pid})};
--- a/src/group/libp2p_group_worker.erl
+++ b/src/group/libp2p_group_worker.erl
@@ -256,14 +256,15 @@ connecting(info, connect_retry_timeout, Data=#data{tid=TID,
                         %% such as when we are gossiped a new peer from an already connected peer
                         %% the dial above would have first performed this connection and then proceeded to dial a stream
                         %% If we dont perform the connect here then we will end up never connecting to gossiped peers with an address with a lower lex order
-                        lager:debug("(~p) peer ~p will not dial a stream to remote peer ~p, but will establish a connection to it if not already done...",[Parent, LocalAddr, MAddr]),
+                        lager:debug("(~p) peer ~p will not dial a stream to remote peer ~p, but will establish a connection to it if not already active...",[Parent, LocalAddr, MAddr]),
                         libp2p_transport:connect_to(MAddr, [], ?CONNECT_TIMEOUT, TID)
                 end
             end),
             {keep_state, stop_connect_retry_timer(Data#data{connect_pid=Pid})};
         true ->
             lager:debug("max connect retries exceeded, going back to targeting"),
-            {next_state, targeting, cancel_connect_retry_timer(Data), ?TRIGGER_TARGETING}
+            libp2p_group_server:clear_target(Data#data.server, Data#data.kind, self(), Data#data.ref),
+            {next_state, targeting, cancel_connect_retry_timer(Data#data{target = undefined}), ?TRIGGER_TARGETING}
     end;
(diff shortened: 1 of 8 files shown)


src/group/libp2p_group_worker.erl
```diff
@@ -231,17 +232,40 @@ connecting(info, connect_retry_timeout, Data=#data{tid=TID,
...
-                case libp2p_swarm:dial_framed_stream(TID, MAddr, Path, M, A) of
-                    {error, Error} ->
-                        Parent ! {connect_error, Error};
-                    {ok, StreamPid} ->
-                        Parent ! {assign_stream, StreamPid}
+                LocalAddr = libp2p_swarm:p2p_address(TID),
+                %% its possible that after exchanging identities two peers will be assigned each other as a target
+                %% and end up dialing each other at pretty much the same time
+                %% resulting in streams being assigned for the first dial, followed by a second set of streams for the second dial
+                %% there is/was then a coin toss ( see handle_assign_stream ) as to which streams win ( we dont want two streams between the same set of peers )
```
Vagabond, line 239: I think it would be better to make the coin toss deterministic based on the identities of the peers involved, then they'd both make the same decision and we'd avoid getting an unlucky coin toss.

andymck: @Vagabond So at the point of dialing ( as far as i can see ) we do not necessarily have the identity of the remote peer, for instance in the case of a new peer we have received over gossip from another connected peer ( which is why i am using the p2p addresses, as these are always available ).  In this scenario, these peers wont have a prior connect attempt, the connect will be performed as part of this initial dialing.

but i am wondering are we overthinking this...If the coin toss is made deterministic then its not a coin toss and if the decision is not based on anything related to the context of the existing or new stream then why use some arbitrary rule to make the decision ?  Why not hardcode the rule to always go with the new stream or stay with the old.  If there is no data or context to influence which option is "best" then why do we care other than both sides need to do the same thing.

One consideration we should keep in mind is, if we decide to drop the existing and go with the new stream then we risk loosing some in flight gossiped msgs.  If we dont care about msg loss here ( i mean a dropped block will be gossiped again, we wont miss a dropped gossiped peer ? and relcast msgs are acked iirc anyway and thus will be resent ) then why not always just accept the new stream on both sides ?

evanmcc:
does this fix supersede #277?

src/group/libp2p_group_worker.erl
Vagabond: You only need to flip a coin if you get a simultaneous dial. Using the local p2p address and the remote p2p address to make the coin flip deterministic would simply be flipping the same coin, rather than both sides flipping coins independently.

andymck: yes, the coin toss does only take place when there is a simultaneous dial ( which is how it was before any changes in this PR )...but flipping the coins independently is where the problem comes in...if both sides flip and get different results they each close a different stream resulting in both old and new streams being closed

Maybe i am not seeing it but why would this be desirable behaviour ?

andymck:
@evanmcc yip, this supersedes 277.  I have now closed that

src/group/libp2p_group_worker.erl
Vagabond: Right, my point is to replace the independent coin flip with a common one. Instead of each side doing rand(2) after a simultaneous dial, they'd instead look at their local identity and the remote entity they're dialing and use that *common* information to decide if they should *both* pick heads or tails.

andymck: gotcha, but keep in mind that dialing can occur before we know the remote peer identity...all we have in some cases is the peerbook entry.  The identities will be swapped as part of the connect performed after dialing is triggered.

But on a general point, what benefit does this coinflip provide ( even if it is a common flip ) ?  It is detached from any context relating to one stream or the other.  Why would we not just say always go with the latest stream or always stick with the existing steam ( on both sides ).  I dont see the value in varying that decision based on the coin flip.

Vagabond: Things like 'latest' or 'existing' are not common on both sides, those are relative values. Things like clock skew or network round trip time can really give both sides a very different view. There needs to be some objective, independent way to break the tie.

When I say Identity, I mean simply the p2p address. That has plenty of entropy and is unique so we can use that to make the decision on which end to chooose.

For example, you could, for a simultaneous connection from a->b and b->a look at if a is greater than b, if it is, you pick the a->b connection (where a is the dialer and b is the dialee) otherwise pick the b->a connedtion, both sides can make this decision independently and come to the same conclusion.

andymck: hmm i am going to think about your last statement further because from my perspective  latest and existing are/were common on both sides.  'existing' is the  currently connected stream from a->b whilst the 'latest' is any new stream which has been established after the 'existing' one connected

but...what may be factoring in here is that the group workers spawn off dialing which i have been thinking is odd given the process is an async worker with no real sync APIs.  Spawning off the dialing however is likely going to exasperate the potential for a worker to get assigned into an undesirable state by dealing with other msgs whist an existing dial is ongoing....not sure what exact implications might be but it smells bad and may be related to this overall problem....

Vagabond: no, in the case of a simultaneous dial you have the following:

a is dialing b
b is dialing a

from a's perspective a->b is the existing connection, and the older one (because it started first from its perspective)
from b's perspetive, b->a is the existing connection, and the older one (because it started first from its perspective)

When a->b connects, b sees it as 'newer' than b->a, but a sees it as older than b->a.

If we simply prefer existing connections, each side would pick the connection with themselves as the dialer, and close the other one, thus both would get closed.

I don't think the async thing is a problem, the problem is that the far end (which may be on the other side of the world) has a different perspective than we do, and is operating independently.

andymck: @Vagabond i see the potential for what you describe.  That said it is not what I was observing with the tests.  With the test runs on buildkite it was simpler.  

a is dialing b
b is dialing a

a's dial completes ( both sides see this as the first/existing connection )
b's dial complete ( both sides see this as the second/new connection )

Both sides toss the coin and get opposite sides resulting in both connections closing.  
If both sides get the same side, all is well.

I'll rework this PR to allow both dials and handle in the coin toss.  Took me a while to see this, thanks for discussing.

Vagabond: Ah yes, on localhost there's usually not enough network latency for the dials to overlap in time. If there's a clear winner we should prefer that and only resort to coin tossing if absolutely necessary.

That said, you don't know what your peer thinks is going on, so if you make a unilateral decision based on your perspective you're likely to be wrong if there's significant latency involved.

andymck:
superseded by https://github.com/helium/erlang-libp2p/pull/281

PR #280, opened 2020-05-17
Experiment to try to use forgetful blooms to dedup gossip

Vagabond:


src/group/libp2p_group_gossip_server.erl
```diff
@@ -185,8 +188,10 @@ handle_cast({send, Key, Fun}, State=#state{}) when is_function(Fun, 0) ->
                   end, Pids),
     {noreply, State};
 
-handle_cast({send, Key, Data}, State=#state{}) ->
+handle_cast({send, Key, Data}, State=#state{seen_messages=Bloom}) ->
     {_, Pids} = lists:unzip(connections(all, State)),
+    %% track we've seen this now
+    forgetful_bloom:set(Bloom, {Key, Data}),
```
andymck, line 194: At this point ( in the gossip server ) we do not know whether the send will be successful.  For example a stream may not yet be associated with the worker

PR #281, opened 2020-05-18
make winner/loser stream close decision deterministic when sim dials occur between peers

andymck:
NOTE: this supersedes https://github.com/helium/erlang-libp2p/pull/278

Related PR ( both issues affect same flow ): https://github.com/helium/erlang-libp2p/pull/282

There is a race condition which can trigger after a libp2p_swarm:connect/2 is called. As part of the connect call the identities of both peers are swapped and peerbook entries are added for the other on each peer.

It is after the peerbook entries are added where the race can occur. Each peer's group workers will request a target from the group server. If they happen to do this at the same time and be assigned each others as the target ( which in the tests can happen frequently ) the workers will dial a stream to each other.

This will result in two sets of streams being created on and for each of the two peers. On each peer, when the second stream comes up it will be assigned to the same group worker as the first. And as there is an existing stream, the group_worker:handle_assign_stream/2 function will flip a coin to determine which stream to keep. As this happens on both peers, should both peers end up with opposite sides of the coin, it will result in both the first and the second streams being closed. The peers are then left without any active streams to each other.

The reconnect logic will then kick in but in many cases we end up repeating the exact same scenario again after which the group worker will give up connecting/dialing and revert back to targeting.

_Observable impact on tests_

The exact extent of the observable behaviour is determined by how closely together the two sets of streams come up ( there could be network latency on one or both routes or a swarm running slowly ). If both peers dials result in streams being assigned in quick succession, the impact is mostly that the tests take longer to advance as they are held up at the _await_gossip_group_ and _await_gossip_streams_ handlers. As the group workers have a back off and will ultimately revert to targeting after they exceed the connect retry limit, the _await*_ functions can timeout leading to test fails.

If one of the two dials take longer for the streams to be generated the impact on the tests is that they advance quickly past the _await*_ functions and start to exchange msgs via gossip using the first stream. During this exchange the second stream can come up ( resulting in both streams being closed due to the peers getting opposite sides of the coin toss ) and the tests fail due to gossip msgs being lost.

**Implemented Solution**
The fix here is to replace the random coin toss with a process that is deterministic on both peers.  This is acheived by both peers choosing a winning stream based on the lexicographic order of their multi addrs.  The dial and associated stream of the peer with the higher ordered peer name will win, the other stream/dial will be closed

_NOTE:_
This solution in not without potential side effects, there remains potential for msgs to be lost.  For example if both peers simultaneously dial but one stream is taking longer to establish.  If, after the first stream is established, msgs start to be exchanged and then the second stream comes up it can result in the currently active stream on which those msgs are being exchanged being closed and msgs being lost mid flight.

### DIFF
--- a/src/group/libp2p_group_worker.erl
+++ b/src/group/libp2p_group_worker.erl
@@ -135,7 +135,7 @@ callback_mode() -> state_functions.
 
 -spec init(Args :: term()) -> gen_statem:init_result(atom()).
 init([Ref, Kind, StreamPid, Server, GroupID, TID, Path]) ->
-    lager:debug("starting group worker of kind: ~p for path: ~p with streamID: ~p",[Kind, Path, StreamPid]),
+    lager:debug("starting group worker of kind ~p and path: ~p with streamID: ~p",[Kind, Path, StreamPid]),
     process_flag(trap_exit, true),
     {ok, connected,
      update_metadata(#data{ref=Ref, tid=TID, server=Server, kind=Kind, group_id=GroupID,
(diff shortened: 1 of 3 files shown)


src/group/libp2p_group_worker.erl
```diff
@@ -362,22 +362,34 @@ closing(EventType, Msg, Data) ->
...
+                #{outbound_stream => CurrentStreamPid, inbound_stream => StreamPid};
+            _ ->
+                %% The existing stream is an inbound stream so the new stream must be an outgoing stream, a local dialer
+                lager:debug("local outbound stream: ~p, inbound: ~p", [StreamPid, CurrentStreamPid]),
+                #{outbound_stream => StreamPid, inbound_stream => CurrentStreamPid}
+        end,
+    %% If local peer's multiaddr is greater lexicographically, then we pick the stream where he is the dialer
+    %% otherwise we pick the other stream which will be where he has received an inbound connection
+    %% NOTE: MAddr here will be the multi addr of the remote peer
+    {WinnerStream, LooserStream} =
```
evanmcc, line 383:
```suggestion
    {WinnerStream, LoserStream} =
```

src/group/libp2p_group_worker.erl
```diff
@@ -362,22 +362,34 @@ closing(EventType, Msg, Data) ->
...
+        case LocalPeerMAddr > MAddr of
+            true ->
+                lager:debug("keeping local outgoing stream and dropping inbound", []),
+                {maps:get(outbound_stream, TaggedStreams), maps:get(inbound_stream, TaggedStreams)};
+            false ->
+                lager:debug("dropping local outgoing stream and switching to inbound", []),
+                {maps:get(inbound_stream, TaggedStreams), maps:get(outbound_stream, TaggedStreams)}
+        end,
+    handle_stream_winner_looser(CurrentStreamPid, WinnerStream, LooserStream, Data).
```
evanmcc, line 392:
```suggestion
    handle_stream_winner_looser(CurrentStreamPid, WinnerStream, LoserStream, Data).
```

src/group/libp2p_group_worker.erl
```diff
@@ -455,6 +467,13 @@ handle_event(EventType, Msg, #data{}) ->
 %% Utilities
 %%
 
+handle_stream_winner_looser(CurrentStream, WinnerStream, LooserStream, Data) when CurrentStream /= WinnerStream->
+    libp2p_framed_stream:close(LooserStream),
+    {ok, update_metadata(Data#data{stream_pid=update_stream(WinnerStream, Data)})};
+handle_stream_winner_looser(_CurrentStream, _WinnerStream, LooserStream, _Data) ->
+    libp2p_framed_stream:close(LooserStream),
```
evanmcc, line 474:
```suggestion
handle_stream_winner_loser(CurrentStream, WinnerStream, LooserStream, Data) when CurrentStream /= WinnerStream->
    libp2p_framed_stream:close(LoserStream),
    {ok, update_metadata(Data#data{stream_pid=update_stream(WinnerStream, Data)})};
handle_stream_winner_loser(_CurrentStream, _WinnerStream, LoserStream, _Data) ->
    libp2p_framed_stream:close(LoserStream),
```

andymck:
@evanmcc spelling corrected,  i thought it looked wrong but got used to it

Vagabond:
Why did we never merge this?

Vagabond (approved):

andymck:
@Vagabond not sure tbh, prob got lost in PR fog

PR #282, opened 2020-05-18
clear target in the group server when worker fails to connect to peer and reverts back to targeting

andymck:


src/group/libp2p_group_relcast_server.erl
```diff
@@ -213,6 +213,18 @@ handle_cast({request_target, Index, WorkerPid, _WorkerRef}, State=#state{tid=TID
                                              {keys, State#state.group_keys}]}},
     libp2p_group_worker:assign_target(WorkerPid, {Target, ClientSpec}),
     {noreply, NewState};
+handle_cast({clear_target, _Kind, _WorkerPid, Ref}, State=#state{workers = Workers}) ->
```
evanmcc, line 216: but the target for each worker in the relcast server is always the same.  maybe we should just drop the cast here?

andymck: Yeah  you are right, relcast doesnt need this but maybe its better to handle the clear target cast as a noop instead, would avoid any future potential confusion if someone is in this flow and wondering why its not handled and doesnt have full context ?

evanmcc: yeah, that's what I meant, sorry.  I think a noop clause here with a good comment is the right thing.

src/group/libp2p_group_gossip_server.erl
```diff
@@ -147,6 +147,18 @@ handle_cast({add_handler, Key, Handler}, State=#state{handlers=Handlers}) ->
     {noreply, State#state{handlers=maps:put(Key, Handler, Handlers)}};
 handle_cast({request_target, inbound, WorkerPid, _Ref}, State=#state{}) ->
     {noreply, stop_inbound_worker(WorkerPid, State)};
+handle_cast({clear_target, _Kind, _WorkerPid, Ref}, State=#state{workers = Workers}) ->
+    lager:debug("clearing target for worker ~p ", [_WorkerPid]),
+    %% the ref is stable across restarts, so use that as the lookup key
+    case lookup_worker(Ref, #worker.ref, State) of
+        Worker=#worker{} ->
```
Vagabond, line 154: In theory the workerpid in here should match the one passed in the message?

src/group/libp2p_group_gossip_server.erl
```diff
@@ -147,6 +147,18 @@ handle_cast({add_handler, Key, Handler}, State=#state{handlers=Handlers}) ->
...
+            %% TODO - should the pid be set to undefined along with target ?
```
Vagabond, line 155: I don't think the pid would be set undefined here, the worker is still running.

src/group/libp2p_group_gossip_server.erl
andymck: yes, Worker.pid() should equal WorkerPid from the message.  As per the comment, the ref is used for the lookup as that is consistent across restarts of the worker

src/group/libp2p_group_gossip_server.erl
andymck: removed todo

PR #285, opened 2020-06-05
framed stream timeouts

evanmcc:
raise the send and recieve timeouts so large messages to slow nodes don't get killed prematurely.


evanmcc:
approved in slack.

PR #287, opened 2020-07-09
Fix old proto syntax

macpie:


src/relay/libp2p_relay_resp.erl
```diff
@@ -68,13 +68,13 @@ error(Req) ->
 -ifdef(TEST).
 
 create_test() ->
-    ?assertEqual(#libp2p_relay_resp_pb{address="123", error=undefined}, create("123")),
```
madninja, line 71: There's an actual code dependency on error being undefined in: https://github.com/helium/erlang-libp2p/blob/master/src/relay/libp2p_stream_relay.erl#L233

which you will need to handle

PR #289, opened 2020-07-15
add twopc test to exercise message loss issues

evanmcc:
This adds a new relcast test with an intentionally fragile two phase commit implementation.

The idea here is to only accept message resends, but halt in the presence of any message that is unsent, either via crashing or by timing out because of lack of progress.

Draft status is because there's still a lot of debug in here.  Of interest are the non-logging changes in server.


src/group/libp2p_group_relcast_server.erl
```diff
@@ -357,7 +360,7 @@ handle_info(force_close, State=#state{}) ->
     %% down the group by exiting the supervisor.
     spawn(fun() ->
                   lager:info("removing group for force_close timeout"),
-                  libp2p_swarm:remove_group(State#state.tid, State#state.group_id)
+                  libp2p_group_mgr:remove_group(State#state.tid, State#state.group_id)
```
evanmcc, line 363: This was a critical change.   This was allowing some bad races where we'd lose state on restarts.

Vagabond: worth just backporting this for now?

evanmcc: yes, I think that we should get just this in ASAP.

PR #290, opened 2020-07-16
port just restart fix

evanmcc:
just the major fix from #289, without all the debugging and new test.

### DIFF
--- a/src/group/libp2p_group_relcast_server.erl
+++ b/src/group/libp2p_group_relcast_server.erl
@@ -357,7 +357,8 @@ handle_info(force_close, State=#state{}) ->
     %% down the group by exiting the supervisor.
     spawn(fun() ->
                   lager:info("removing group for force_close timeout"),
-                  libp2p_group_mgr:remove_group(State#state.tid, State#state.group_id)
+                  Mgr = libp2p_group_mgr:remove_group(State#state.tid),
+                  libp2p_group_mgr:remove_group(Mgr, State#state.group_id)
           end),
     {noreply, State#state{close_state=closing}};
 handle_info(inbound_tick, State = #state{store=Store}) ->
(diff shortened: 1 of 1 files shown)


src/group/libp2p_group_relcast_server.erl
```diff
@@ -357,7 +357,7 @@ handle_info(force_close, State=#state{}) ->
     %% down the group by exiting the supervisor.
     spawn(fun() ->
                   lager:info("removing group for force_close timeout"),
-                  libp2p_swarm:remove_group(State#state.tid, State#state.group_id)
+                  libp2p_group_mgr:remove_group(State#state.tid, State#state.group_id)
```
Vagabond, line 360: Why is this needed, given that libp2p_swarm:remove_group calls libp2p_group_mgr:remove_group?

Vagabond: Actually, this calls via the TID, but libp2p_group_mgr:remove_group expects a pid as the first argument?

src/group/libp2p_group_relcast_server.erl
```diff
@@ -357,7 +357,8 @@ handle_info(force_close, State=#state{}) ->
     %% down the group by exiting the supervisor.
     spawn(fun() ->
                   lager:info("removing group for force_close timeout"),
-                  libp2p_swarm:remove_group(State#state.tid, State#state.group_id)
+                  Mgr = libp2p_group_mgr:remove_group(State#state.tid),
```
Vagabond, line 360: D you mean `libp2p_group_mgr:mgr` here?

evanmcc: lol.  So I reverted the change here totally, and it broke again as expected.  trying with the fix you suggested, I suspect that it will also break.

UPDATE: before I finished typing the above sentence, the "corrected" code caused the same crash.

Vagabond: I think it worked because it was failing to close.

PR #291, opened 2020-07-16
Add max_tcp_connections

macpie:


src/libp2p_transport_tcp.erl
```diff
@@ -68,6 +68,7 @@
          negotiated_nat=false :: boolean()
         }).
 
+-define(DEFAULT_MAX_TCP_CONNECTIONS, 1024).
```
evanmcc, line 71: make this tunable, please?

evanmcc: nm, ignore me, I didn't read it right.

macpie: (just in case) 1024 is ranch's default already

PR #293, opened 2020-07-30
stop gossiping connected peers for scalability reasons

evanmcc:
change the code so that we don't use it anywhere internally, as well.

### DIFF
--- a/src/peerbook/libp2p_peerbook.erl
+++ b/src/peerbook/libp2p_peerbook.erl
@@ -8,6 +8,7 @@
          join_notify/2, changed_listener/1, update_nat_type/2,
          register_session/3, unregister_session/2,
          blacklist_listen_addr/3,
+         sessions/1,
          add_association/3, lookup_association/3]).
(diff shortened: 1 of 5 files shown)


src/peerbook/libp2p_peerbook.erl
```diff
@@ -475,10 +482,7 @@ handle_info(notify_timeout, State=#state{}) ->
 handle_info(gossip_peers_timeout, State=#state{peerbook=Handle, tid=TID}) ->
     %% A peer is eligilble for gossiping if it is dialable and has a
     %% minimum number of outbound connections.
-    IsEligibleGossipPeer = fun(Peer) ->
```
evanmcc, line 478: this is a potentially questionable decision.  If it's not a bad one, mabye we should use `random` here instead of `fold_peers`?

madninja: Yeah.. by not knowing how connected the peer is you could be picking some really bad counterparts which would not help gossip.

PR #295, opened 2020-08-03
limit peer connection gossip, use random for eligible peers

evanmcc:
I think that this is a better version of #293.

Instead of dropping our connections, we limit the number of them we keep to a smaller number.

Also, instead of keeping an unscalable and seldom-updated bunch of state hanging around, we use random instead to select gossip peers.

### DIFF
--- a/src/peerbook/libp2p_peer.erl
+++ b/src/peerbook/libp2p_peer.erl
@@ -49,7 +49,7 @@ from_map(Map, SigFun) ->
                                {Type, #libp2p_association_list_pb{associations=AssocEntries}}
                        end, maps:get(associations, Map, [])),
     Connected0 = maps:get(connected, Map, []),
-    MaxConns = application:get_env(libp2p, max_connected_peers, 8),
+    MaxConns = application:get_env(libp2p, max_peers_to_gossip, 8),
     Connected =
         case length(Connected0) =< MaxConns of
             true -> Connected0;
--- a/src/peerbook/libp2p_peer.erl
+++ b/src/peerbook/libp2p_peer.erl
@@ -48,9 +48,16 @@ from_map(Map, SigFun) ->
     Assocs = lists:map(fun({Type, AssocEntries}) ->
                                {Type, #libp2p_association_list_pb{associations=AssocEntries}}
                        end, maps:get(associations, Map, [])),
+    Connected0 = maps:get(connected, Map, []),
+    MaxConns = application:get_env(libp2p, max_connected_peers, 8),
+    Connected =
+        case length(Connected0) =< MaxConns of
+            true -> Connected0;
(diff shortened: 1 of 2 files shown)


Vagabond:
Does making the connected peer list this random interact badly with the 'is newer' peer comparisons?

src/peerbook/libp2p_peerbook.erl
```diff
@@ -505,6 +458,11 @@ terminate(_Reason, _State) ->
 %% Internal
 %%
 
+eligible_gossip_peer(Peer) ->
```
Vagabond, line 461: So the goal here is to deliver some dialable and reasonably well connected peers from the peerbook?

evanmcc: the idea was to replicate the existing functionality without tick and without storing stuff in the state no fold, etc).

src/peerbook/libp2p_peer.erl
```diff
@@ -48,9 +48,16 @@ from_map(Map, SigFun) ->
     Assocs = lists:map(fun({Type, AssocEntries}) ->
                                {Type, #libp2p_association_list_pb{associations=AssocEntries}}
                        end, maps:get(associations, Map, [])),
+    Connected0 = maps:get(connected, Map, []),
+    MaxConns = application:get_env(libp2p, max_connected_peers, 8),
```
madninja, line 52: This is not really limiting the number of connected peers is it? Naming is hard :-/ Maybe call it max_enoded_peers or max_gossiped_peer_connnections

src/peerbook/libp2p_peerbook.erl
```diff
@@ -340,32 +335,12 @@ handle_gossip_data(_StreamPid, DecodedList, Handle) ->
...
-                                 end
-                         end,
-            %% Try 10 times to pick a random peer that is actually
-            %% available in the peerbook
-            case SelectPeer(10) of
-                {ok, Peer} -> {send, libp2p_peer:encode_list([Peer])};
-                {error, not_found} -> ok
-            end
+init_gossip_data(Peerbook) ->
+    case random(Peerbook, [], fun eligible_gossip_peer/1) of
```
madninja, line 339: It'll be interesting to see how this behaves in the field. I.E. it could return non eligible peers a long time which could affect gossip.

evanmcc: It'll never return a non-eligible peer, it'll just give up if it doesn't find one after fifteen tries.  I'll increase the number of retries so that it effectively will always find one.

evanmcc:
@Vagabond we removed connected_peers from the `supersedes` function a while ago, I don't think that anything internal to the peerbook uses that information.

PR #306, opened 2021-01-05
fdset_loop monitors parent to clean itself if parent goes down

macpie:


src/libp2p_transport_tcp.erl
```diff
@@ -218,7 +223,8 @@ fdset(#tcp_state{socket=Socket}=State) ->
     Parent = self(),
     case erlang:get(fdset_pid) of
         undefined ->
-            Pid = spawn(fun() ->
+            Pid = erlang:spawn(fun() ->
+                erlang:monitor(process, Parent),
```
madninja, line 227: Mind indenting this right?

madninja (changes requested):
Just an indentation request for that monitor statement. Otherwise we know this works now

PR #309, opened 2021-01-26
Don't observe addresses when we have been given one with NAT

Vagabond:


evanmcc (approved):

evanmcc:
We tested this overnight, and on nodes that have the issue to the degree that they generate many leaked IPs per day, and they stayed at one IP, while maintaining their ability to work as a relay.

PR #311, opened 2021-01-28
Avoid unnecessary pid/tid conversions when dialing

Vagabond:
This seems to add a dialyzer error that I do not have time to chase down.


evanmcc (commented):
merge the other one into and lets see if it goes green?  let me know if you need a hand with dialyzer.

evanmcc (approved):

PR #312, opened 2021-01-28
when doing NAT, make sure the transport knows about nat server crashes

evanmcc:
we've seen cases where the nat server crashes from not being able to renew and then tcp transport never retries establishing a nat connection.

### DIFF
--- a/src/libp2p_transport_tcp.erl
+++ b/src/libp2p_transport_tcp.erl
@@ -544,7 +544,8 @@ handle_info({nat_discovered, InternalAddr, ExternalAddr}, State=#state{tid=TID})
     end;
 handle_info(no_nat, State) ->
     {noreply, State#state{negotiated_nat = false}};
-handle_info({'DOWN', _Ref, process, Pid, _Reason}, State=#state{nat_server = {_NatRef, NatPid}}) when Pid == NatPid ->
+handle_info({'DOWN', Ref, process, Pid, _Reason}, State=#state{nat_server = {NatRef, NatPid}})
+  when Pid == NatPid andalso Ref == NatRef ->
     {noreply, State#state{nat_server = undefined, negotiated_nat = false}};
 handle_info({'DOWN', _Ref, process, Pid, Reason}, State=#state{tid=TID}) when Reason /= normal; Reason /= shutdown ->
     %% check if this is a listen socket pid
--- a/src/libp2p_transport_tcp.erl
+++ b/src/libp2p_transport_tcp.erl
@@ -542,6 +542,8 @@ handle_info({nat_discovered, InternalAddr, ExternalAddr}, State=#state{tid=TID})
             lager:warning("no listener detected for ~p", [InternalAddr]),
             {noreply, State}
     end;
+handle_info(no_nat, State) ->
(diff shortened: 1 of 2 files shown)


src/nat/libp2p_nat_server.erl
```diff
@@ -116,7 +116,7 @@ handle_info(renew, #state{tid=TID, transport_tcp=Pid, internal_address=IntAddr,
         {error, _Reason} ->
             ok = remove_multi_addr(TID, ExtAddr0, ExtPort0),
             lager:warning("failed to renew lease for port ~p: ~p", [{IntPort, ExtPort0}, _Reason]),
-            {stop, renew_failed}
+            {stop, renew_failed, State}
```
Vagabond, line 119: I would propose we retry renewing later rather than stopping here because the router's upnp/pmp server often will restart.

Vagabond: Unless stopping will start a new server that will retry itself? we *know* there was a nat server available at one point.

evanmcc: this is the nat server that's stopping, not the transport.  This is just to fix the crash, hopefully with the return fixed it won't spew quite so badly.  it'll restart, but I am not sure that it'll do anything without more work on the transport side.  I'm not sure if the clearing of the state that happens in the other module is enough or if some discovery process needs to be kicked back off on the transport side.

src/nat/libp2p_nat_server.erl
```diff
@@ -114,9 +114,11 @@ handle_info(renew, #state{tid=TID, transport_tcp=Pid, internal_address=IntAddr,
             end,
             {noreply, State#state{external_address=ExtAddr1, external_port=ExtPort1, lease=Lease, since=Since}};
         {error, _Reason} ->
+            %% this might lead to a period of undialability
             ok = remove_multi_addr(TID, ExtAddr0, ExtPort0),
             lager:warning("failed to renew lease for port ~p: ~p", [{IntPort, ExtPort0}, _Reason]),
-            {stop, renew_failed}
+            _ = erlang:send_after(5000, self(), renew),
```
Vagabond, line 120: this is likely far too short, would suggest at least 5 minutes AND telling the transport_tcp to set `negotiated_nat = false` in the meantime.

src/libp2p_transport_tcp.erl
```diff
@@ -529,11 +530,22 @@ handle_info({nat_discovered, InternalAddr, ExternalAddr}, State=#state{tid=TID})
...
+            {noreply, State#state{negotiated_nat = true,
+                                  observed_addrs = sets:new(),
+                                  nat_server = {Ref, Pid}}};
         _ ->
+            lager:warning("no listener detected for ~p", [InternalAddr]),
             {noreply, State}
     end;
+handle_info(no_nat, State) ->
+    {noreply, State#state{negotiated_nat = false}};
+handle_info({'DOWN', _Ref, process, Pid, _Reason}, State=#state{nat_server = {_NatRef, NatPid}}) when Pid == NatPid ->
```
Vagabond, line 547: Match the Ref here too?

Vagabond (approved):

PR #313, opened 2021-02-04
Do not try infinite leases on NAT-PMP, lease time of 0 means delete

Vagabond:


macpie (approved):

src/nat/libp2p_nat.erl
```diff
@@ -185,25 +185,29 @@ add_port_mapping(Context, InternalPort, ExternalPort0, Retry) ->
 %% @doc
 %% @end
 %%--------------------------------------------------------------------
--spec retry_matrix(integer()) -> list().
-retry_matrix(Port) ->
+-spec retry_matrix(nat:nat_ctx(), integer()) -> list().
```
vihu, line 188: This breaks the `retry_matrix_test` eunit.

cc: @madninja

PR #316, opened 2021-03-02
Store peerbook address keys reversed for better random distribution

Vagabond:
Storing peer address keys with the type bytes at the start leads to
extremely bad distribution of the keys which makes the random function
very likely to select the same peers over and over.

This patch stores the keys reversed so the tag bytes are the least
significant ones. The peerbook is rewritten on startup if needed.

### DIFF
--- a/src/peerbook/libp2p_peerbook.erl
+++ b/src/peerbook/libp2p_peerbook.erl
@@ -169,7 +169,8 @@ random(Peerbook, Exclude) ->
 random(Peerbook, Exclude, Pred) ->
     random(Peerbook, Exclude, Pred, 15).
 
-random(Peerbook=#peerbook{store=Store, stale_time=StaleTime}, Exclude, Pred, Tries) ->
+random(Peerbook=#peerbook{store=Store, stale_time=StaleTime}, Exclude0, Pred, Tries) ->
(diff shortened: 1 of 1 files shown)


src/peerbook/libp2p_peerbook.erl
```diff
@@ -185,7 +185,7 @@ random(Peerbook=#peerbook{store=Store, stale_time=StaleTime}, Exclude, Pred, Tri
                 false ->
```
evanmcc, line 185: map rev over the exclude list or it won't match right, I think

evanmcc (commented):
other than the one comment this looks good.  I'm concerned about the rewrite at the beginning, I'd rather just drop it, but this should work too.

src/peerbook/libp2p_peerbook.erl
Vagabond: good catch!

Vagabond: done

Vagabond:
> other than the one comment this looks good. I'm concerned about the rewrite at the beginning, I'd rather just drop it, but this should work too.

I tested the rewrite on 1gb ram hotspots and it went fine.

madninja (approved):

evanmcc (approved):

PR #330, opened 2021-04-08
Fix libp2p_session:close when noproc

macpie:


madninja (approved):
I think this fits with what this function was intended to achieve

Vagabond (approved):

PR #337, opened 2021-05-13
Use DNS to lookup seed nodes

mrallen1:
Problem to solve: We are currently publishing seed node data in a TXT record which has a maximum length. If we were to publish seed node records as a list of A records linked to a single name, we could more easily update DNS information about seed nodes, for example, adding to the pool of available seeds beyond the length of the current TXT record.

Proposed solution: Lookup DNS names at the point of seed node assignment (to bootstrap peers) and fall back to choosing one of the records from config data or the TXT record inherited from miner's DNS lookup.

### DIFF
--- a/src/group/libp2p_group_gossip_server.erl
+++ b/src/group/libp2p_group_gossip_server.erl
@@ -4,6 +4,7 @@
 -behavior(libp2p_gossip_stream).
 
 -include("gossip.hrl").
+-include_lib("kernel/include/inet.hrl"). %% for DNS lookup
 
 %% API
 -export([start_link/2]).
@@ -39,6 +40,10 @@
 -define(DEFAULT_MAX_INBOUND_CONNECTIONS, 10).
 -define(DEFAULT_DROP_TIMEOUT, 5 * 60 * 1000).
 -define(GROUP_ID, "gossip").
+-define(DNS_RETRIES, 3).
+-define(DNS_TIMEOUT, 2000). % millis
+-define(DNS_SLEEP, 100). % millis
+-define(DEFAULT_SEED_DNS_PORTS, [2154, 443]).
 
 %% API
 %%
@@ -190,8 +195,9 @@ handle_cast({request_target, seed, WorkerPid, Ref}, State=#state{tid=TID, seed_n
     LocalAddr = libp2p_swarm:p2p_address(TID),
     %% Exclude the local swarm address from the available addresses
     ExcludedAddrs = CurrentAddrs ++ [LocalAddr],
-    TargetAddrs = sets:to_list(sets:subtract(sets:from_list(SeedAddrs),
+    BaseAddrs = sets:to_list(sets:subtract(sets:from_list(SeedAddrs),
                                              sets:from_list(ExcludedAddrs))),
+    TargetAddrs = maybe_lookup_seed_in_dns(BaseAddrs),
     {noreply, assign_target(WorkerPid, Ref, TargetAddrs, State)};
 handle_cast({send, Key, Fun}, State=#state{}) when is_function(Fun, 0) ->
     %% use a fun to generate the send data for each gossip peer
@@ -361,7 +367,7 @@ assign_target(WorkerPid, WorkerRef, TargetAddrs, State=#state{workers=Workers, s
                     State
             end;
         _ ->
-            SelectedAddr = mk_multiaddr(lists:nth(rand:uniform(length(TargetAddrs)), TargetAddrs)),
+            SelectedAddr = mk_multiaddr(choose_random_element(TargetAddrs)),
             ClientSpec = {SupportedPaths, {libp2p_gossip_stream, [?MODULE, self(), Bloom]}},
             libp2p_group_worker:assign_target(WorkerPid, {SelectedAddr, ClientSpec}),
             %% the ref is stable across restarts, so use that as the lookup key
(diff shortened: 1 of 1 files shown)


src/group/libp2p_group_gossip_server.erl
```diff
@@ -335,7 +340,7 @@ assign_target(WorkerPid, WorkerRef, TargetAddrs, State=#state{workers=Workers, s
                     State
             end;
         _ ->
-            SelectedAddr = mk_multiaddr(lists:nth(rand:uniform(length(TargetAddrs)), TargetAddrs)),
+            SelectedAddr = mk_multiaddr(maybe_lookup_seed_in_dns(TargetAddrs)),
```
madninja, line 343: Isn't this function used to assign both seed and non-seed targets? If so, this is probably not doing the right thing?

mrallen1: I'm not sure? It seems to be used by worker type `seed` only so that's why I thought this is the place to modify

Vagabond: This should be done here: https://github.com/helium/erlang-libp2p/blob/f4d1d84c8f312555da8dbf3df6c41e064abd2fdb/src/group/libp2p_group_gossip_server.erl#L184 instead (just lookup a DNS seed node here and pass a 1 element list to asign_target).

src/group/libp2p_group_gossip_server.erl
```diff
@@ -351,6 +357,62 @@ assign_target(WorkerPid, WorkerRef, TargetAddrs, State=#state{workers=Workers, s
...
+    case inet_res:gethostbyname(Name, inet, ?DNS_TIMEOUT) of
+        {error, _} = Error ->
+            lager:debug("attempt ~p: got ~p", [Attempts, Error]),
+            timer:sleep(?DNS_SLEEP),
+            attempt_dns_lookup(Name, Attempts - 1);
+        {ok, HostRec} -> {ok, HostRec}
+    end.
+
+convert_dns_record(#hostent{h_addr_list = AddrList}) ->
+    IPStr = inet:ntoa(choose_random_element(AddrList)),
```
Vagabond, line 407: Arguably you could resolve all the addresses here, vs choosing one early.

src/group/libp2p_group_gossip_server.erl
```diff
@@ -351,6 +357,62 @@ assign_target(WorkerPid, WorkerRef, TargetAddrs, State=#state{workers=Workers, s
...
+        CName ->
+            case attempt_dns_lookup(CName, ?DNS_RETRIES) of
+                {error, _} = Error ->
+                    lager:error("DNS lookup of ~p resulted in ~p; falling back", [CName, Error]),
+                    choose_random_element(TargetAddrs);
+                {ok, DNSRecord} ->
+                    lager:debug("successful DNS lookup result: ~p", [lager:pr(DNSRecord, inet)]),
+                    %% to help migitate possible eclipse attacks we will blend the DNS result
+                    %% with the static list of seed nodes and randomly pick one of the results
+                    choose_random_element([convert_dns_record(DNSRecord) | TargetAddrs])
```
Vagabond, line 392: Return the whole list here and let the choose_random_element in assign_target handle it?

src/group/libp2p_group_gossip_server.erl
```diff
@@ -377,6 +383,62 @@ assign_target(WorkerPid, WorkerRef, TargetAddrs, State=#state{workers=Workers, s
...
+
+convert_dns_records(#hostent{h_addr_list = AddrList}) ->
+    [ format_dns_addr(A) || A <- AddrList ].
+
+format_dns_addr(Addr) ->
+    IPStr = inet:ntoa(Addr),
+    PortList = application:get_env(libp2p, seed_dns_ports, ?DEFAULT_SEED_DNS_PORTS),
+    PortStr = to_string(choose_random_element(PortList)),
+    %% XXX this can potentially return `error(badarg)' so maybe try/catch?
+    libp2p_crypto:p2p_to_pubkey_bin("/ip4/" ++ IPStr ++ "/tcp/" ++ PortStr).
```
Vagabond, line 436: This is not what this function is for, you should just return the concatenated string here?

Vagabond, line 436:
```suggestion
    "/ip4/" ++ IPStr ++ "/tcp/" ++ PortStr.
```

mrallen1: updated

src/group/libp2p_group_gossip_server.erl
```diff
@@ -377,6 +383,61 @@ assign_target(WorkerPid, WorkerRef, TargetAddrs, State=#state{workers=Workers, s
...
+    case application:get_env(libp2p, use_dns_for_seeds, false) of
+        false ->
+            TargetAddrs;
+        true ->
+            lookup_seed_from_dns(TargetAddrs)
+    end.
+
+choose_random_element([E]) -> E;
+choose_random_element(L) when is_list(L) ->
+    lists:nth(rand:uniform(length(L), L)).
```
Vagabond, line 396:
```suggestion
    lists:nth(rand:uniform(length(L)), L).
```

Vagabond, line 396:
```suggestion
    lists:nth(rand:uniform(length(L)), L).
```

PR #343, opened 2021-05-21
Fix behavior(gen_server)

rsnodgrass:


madninja:
Were you actually seeing an error there? Most erlang versions will recognize both the UK and US spellings of that word

rsnodgrass:
No problem. Was being consistent with other uses in the repo. Feel free to ignore.

PR #345, opened 2021-05-29
Do some more work upfront to speed up gc

michaeldjeffrey:
I profiled this by writing a bunch of records that look like what libp2p_config writes to ets, using References instead of Pids and randomizing the contents a bit.

```
> match_delete_gc().

129,748 rows

Running gc
.  Found 19,733 records marked for delete
.  Match Delete right hand side: 176 seconds
.  Match Delete add_info: 0 seconds
GC finished in 176 seconds
Table shrank by 78,886 rows
Before: 12,9750
After: 50,864

> spec_select_gc().

10,142,507 rows

Running gc
.  Collecting Keys: 4 seconds
.  Collecting Pids: 1,365,358 in 2 seconds
.  Collecting AddrInfos: 1,365,358 in 92 seconds
.  Deleting Pids: 17 seconds
.  Deleting AddrInfos: 1 seconds
GC finished in 118 seconds

Table shrank by 8,197,425 rows
Before: 10,259,171
After: 2,061,746
```

Collecting the AddrInfo records takes its sweet time because I'm checking them against the marked Pids, is there a way to speed that up?

My hope is that the table won't normally make it to 10mil rows. 
When I checked, there was about 1k `{delete, Pid}` added every minute.
If we keep the normal 5 minute GC period, we should be able to keep up.

### DIFF
--- a/src/libp2p_config.erl
+++ b/src/libp2p_config.erl
@@ -145,11 +145,27 @@ lookup_handlers(TID, TableKey) ->
 
 
 gc_pids(TID) ->
-    ToDelete = ets:match_object(TID, {{delete, '_'}, '_'}),
-    lists:foreach(fun({{delete, Pid}, Pid}) ->
-                  ets:match_delete(TID, {'_', Pid}),
-                  ets:match_delete(TID, {{?ADDR_INFO, '_', Pid}, '_'})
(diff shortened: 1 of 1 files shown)


mrallen1 (approved):
:+1:

PR #346, opened 2021-06-03
Re-work ETS GC to match all the keys it needs to and be iterative

Vagabond:


Vagabond:
This could be made 1-pass by using an ordered set and making the delete keys sort before anything else and then accumulating delete pids in the loop.

Vagabond:
oh, it IS an ordered set.

src/libp2p_config.erl
```diff
@@ -119,20 +120,15 @@ remove_pid(TID, Kind, Ref) ->
 
 -spec remove_pid(ets:tab(), pid()) -> true.
 remove_pid(TID, Pid) ->
-    ets:insert(TID, {{delete, Pid}, Pid}),
-    %spawn(fun() ->
-    %              ets:match_delete(TID, {'_', Pid}),
-    %              ets:match_delete(TID, {{?ADDR_INFO, '_', Pid}, '_'}),
-    %              ok
-    %      end),
+    ets:insert(TID, {{?DELETE, Pid}, Pid}),
```
mrallen1, line 123: Shouldn't we let the ets call return `true` here and make the spec allow for a badarg result?

src/libp2p_config.erl
```diff
@@ -145,27 +141,44 @@ lookup_handlers(TID, TableKey) ->
...
+
+    %% because this table is an ordered set, we can traverse in a known order and we know we'll see the delete keys first
+    %% so we don't need to explicitly find them beforehand
+    {Matches, Continuation} = ets:select(TID, ets:fun2ms(fun({K, _}) when element(1, K) == addr_info -> K end) ++
+                                         ets:fun2ms(fun({_, V}=O) when is_pid(V) -> O end), 100),
+
+    gc_loop(Matches, Continuation, TID, sets:new()).
+
+gc_loop([], '$end_of_table', TID, _) ->
+    %% indicate we're doing doing a destructive iteration
```
Vagabond, line 155:
```suggestion
    %% indicate we're done doing a destructive iteration
```

mrallen1 (approved):
lgtm - just the one comment

PR #347, opened 2021-06-09
Handle end being reached after eating all items

michaeldjeffrey:
Forward to OG end of table clause.

### DIFF
--- a/src/libp2p_config.erl
+++ b/src/libp2p_config.erl
@@ -159,7 +159,11 @@ gc_loop([], '$end_of_table', TID, _) ->
     ets:safe_fixtable(TID, false),
     ok;
 gc_loop([], Continuation, TID, Pids) ->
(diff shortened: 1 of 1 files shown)


src/libp2p_config.erl
```diff
@@ -159,7 +159,11 @@ gc_loop([], '$end_of_table', TID, _) ->
     ets:safe_fixtable(TID, false),
     ok;
 gc_loop([], Continuation, TID, Pids) ->
-    {Matches, NewContinuation} = ets:select(Continuation),
+    {Matches, NewContinuation} =
+        case ets:select(Continuation) of
+            '$end_of_table' = End -> {[], End};
```
evanmcc, line 164: won't this go into an infinite loop?

evanmcc: oh, no, OK

evanmcc (approved):

PR #362, opened 2021-07-01
Clean-up corrupted cache file

xandkar:
fixes helium/miner#792

### DIFF
--- a/src/libp2p_cache.erl
+++ b/src/libp2p_cache.erl
@@ -74,8 +74,22 @@ init([TID]) ->
     libp2p_swarm_auxiliary_sup:register_cache(TID),
     SwarmName = libp2p_swarm:name(TID),
     DataDir = libp2p_config:base_dir(TID),
-    Opts = [{file, filename:join([DataDir, SwarmName, "cache.dets"])}],
-    {ok, Dets} = dets:open_file(SwarmName, Opts),
+    CacheFilePath = filename:join([DataDir, SwarmName, "cache.dets"]),
+    Open = fun() -> dets:open_file(SwarmName, [{file, CacheFilePath}]) end,
+    {ok, Dets} =
+        case Open() of
+            {error, {not_a_dets_file, CacheFilePath}} ->
+                lager:error(
+                    "DETS file corrupted: ~p. Deleting it.",
+                    [CacheFilePath]
+                ),
+                ok = file:delete(CacheFilePath),
+                %% A single retry should be sufficient, if the problem is
+                %% indeed simply a corrupted file.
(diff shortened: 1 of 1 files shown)


src/libp2p_cache.erl
```diff
@@ -74,8 +74,18 @@ init([TID]) ->
...
     SwarmName = libp2p_swarm:name(TID),
     DataDir = libp2p_config:base_dir(TID),
-    Opts = [{file, filename:join([DataDir, SwarmName, "cache.dets"])}],
-    {ok, Dets} = dets:open_file(SwarmName, Opts),
+    CacheFilePath = filename:join([DataDir, SwarmName, "cache.dets"]),
+    Opts = [{file, CacheFilePath}],
+    Open = fun() -> dets:open_file(SwarmName, Opts) end,
+    {ok, Dets} =
+        case Open() of
+            {error, {not_a_dets_file, CacheFilePath}} ->
```
Vagabond, line 82: Would suggest making this `{error, _}`

src/libp2p_cache.erl
```diff
@@ -74,8 +74,18 @@ init([TID]) ->
...
+                %% Correupted file. Single retry should be fine.
```
Vagabond, line 83:
```suggestion
                %% Corrupted file. Single retry should be fine.
```

src/libp2p_cache.erl
xandkar: I'd hesitate to do a specific thing for a general case, but I guess this just a cache file, so the worst scenario isn't that bad :man_shrugging:

src/libp2p_cache.erl
```diff
@@ -74,8 +74,22 @@ init([TID]) ->
...
+    Open = fun() -> dets:open_file(SwarmName, Opts) end,
+    {ok, Dets} =
+        case Open() of
+            {error, Reason} ->
+                %% Likely corrupted file. Single retry should be fine.
+                lager:warning(
+                    "Failed to open DETS file: ~p, for reason: ~p. Deleting.",
+                    [CacheFilePath, Reason]
+                ),
+                ok = file:delete(CacheFilePath),
```
xandkar, line 88: @Vagabond On the second thought - this is bad, since if we get the error for some reason other than corrupt _existing_ file, the deletion will fail, so we need to either go back to handling specific `dets:open_file` error or/and handle|ignore output of `file:delete`.

Vagabond: ok, I suppose it will badmatch in a new and exciting way

Vagabond (approved):

PR #365, opened 2021-07-20
Add bloom filter to dedup gossip

Vagabond:


evanmcc (approved):
This seems to work really well.

Issue #367, opened 2021-07-29
Treat rfc6598 addresses (100.64.0.0/10) as non-public


PaulVMo:
As observed by the community of validator operators, there are several validators and hotspots in the peer book with IP addresses in the 100.64.0.0/10 address. These are non-routable addresses in the shared address space as defined by RFC6598.

The [is_public()](https://github.com/helium/erlang-libp2p/blob/0b8f41be95156d94746a53fd62514f95d0d66f58/src/libp2p_transport_tcp.erl#L374) check for IP addresses currently only considered RFC1918 address (10.0.0.0/8, 172.16.0.0/12, and 192.168.0.0). However there are several other non-routable IP addresses ranges. RFC6598 seems to be the next most common set apart from

kellybyrd:
There are more non-public ranges than just RFC1918 and RFC6598. Many firewalls solve this problem by downloading a [bogon](https://team-cymru.com/community-services/bogon-reference/) list. The list includes RFC1918, RFC6598, the link local addresses, various test ranges, etc.  It also includes ranges that have not been assigned to an ISP or end user yet. The lists are updated periodically, typically once a week or something like this. 

IMO, the minimum would be to use a static copy of the "[The Text Bogon List, Aggregated](https://www.team-cymru.org/Services/Bogons/bogon-bn-agg.txt)"  from  copied from here:  https://team-cymru.com/community-services/bogon-reference/bogon-reference-http/ 

Even better would be to build a mechanism to update that list periodically (1-2 a day at most should be fine, I think my firewall is set to update once a week) and use the "IPv4 Fullbogons" list on that same bogon reference page.

Vagabond:
That's a pretty short list, how often does it realistically change?

kellybyrd:
> That's a pretty short list, how often does it realistically change?

The version of the list he used in the code is just the RFC'd ranges and in ipv4 land, that's really all there is anymore, everything else is allocated. So it's only likely to change when an existing range is changed. So maybe once every few years? But if there is ever ipv6 support than that changes more often.

PaulVMo:
To add, more extensive bogon lists also includes allocated public IPs address blocks that have not been assigned to an ISP. I do not think that is necessary as the larger risk is misconfiguration using private/shared blocks rather than the broader set of allocated but yet to be assigned IPs. Someone would need to intentionally use an unassigned IP which is not really to their advantage in any way. Misconfigure is the more likely scenario.

[...]

PR #368, opened 2021-07-29
Expand IP bogon list to include 100.64.0.0/10 and others

PaulVMo:
Update to transport_tcp to classify additional IPv4 address blocks as non-public using a defined list of known bogons (bogus or non-publicly routable IPs). This addresses issue #367 to prevent validators and hotspots from publishing unreachable address in the peerbook.


src/libp2p_transport.erl
```diff
@@ -59,8 +59,8 @@ for_addr(TID, Addr) ->
 %%--------------------------------------------------------------------
 %% @doc
 %% Addresses are sorted by priority ranking from 1 to 5 (1 highest and 5 lowest priority)
-%% 1 = NON rfc1918 IP (public IPs), 2 = P2P circuit address (relay), 3 = P2P address,
-%% 4 = rfc1918 IPs (private/local IPs), 5 = Proxy transport
+%% 1 = Non-Private IP (public IPs), 2 = P2P circuit address (relay), 3 = P2P address,
+%% 4 = Public IPs (shared/reserved/private/local IPs), 5 = Proxy transport
```
kellybyrd, line 63: Don't you mean *Private* IPs instead of `Public IPs` here? 

Nit: Actually these things technically aren't all private. Some are multicast or reserved. Personally, I think using the term bogon in code will lead future devs to look up what that means and understand.  This "private vs bogon" comment applies to this whole PR.

src/libp2p_transport_tcp.erl
```diff
@@ -31,6 +31,23 @@
 
 -type opt() :: {listen, [listen_opt()]}.
 
+%% list of non-publicly routable IP address blocks in CIDR notation: {IP, bits in mask}
+%% exclude 0.0.0.0/8 as this is required for the relay test suite
+-define(BOGONS, [{{10, 0, 0, 0}, 8},
```
kellybyrd, line 36: nit: With lists like this,  I think a comment with a link to the source you pulled it from is useful.

kellybyrd (commented):
I only validated the list and made comments on the use of "private". Others should validate the actual code that applies the list is correct.

src/libp2p_transport_tcp.erl
PaulVMo: Thanks. Will add the source.

src/libp2p_transport.erl
PaulVMo: Good catch on the Public when I meant private. My use of private_ip was to make it a little more obvious but fair point on it not just being private IPs. I will play with some other names for the function.

src/libp2p_transport_tcp.erl
```diff
@@ -1033,19 +1050,21 @@ do_identify(Session, Identify, State=#state{tid=TID}) ->
...
+bogon_ip_mask_test() ->
+    ?assertEqual(8, bogon_ip_mask({10, 0, 0, 0})),
+    ?assertEqual(8, bogon_ip_mask({10, 20, 0, 0})),
+    ?assertEqual(8, bogon_ip_mask({10, 1, 1, 1})),
+    ?assertEqual(16, bogon_ip_mask({192, 168, 10, 1})),
+    ?assertEqual(16, bogon_ip_mask({192, 168, 20, 1})),
+    ?assertEqual(16, bogon_ip_mask({192, 168, 30, 1})),
+    ?assertEqual(12, bogon_ip_mask({172, 16, 1, 0})),
+    ?assertEqual(12, bogon_ip_mask({172, 16, 10, 0})),
+    ?assertEqual(12, bogon_ip_mask({172, 16, 100, 0})),
```
Vagabond, line 1062: Mind testing some of the new prefixes?

Vagabond:
Adding 127.0.0.1 breaks the nat_SUITE server test, can you do something like this:

```
--- a/test/nat_SUITE.erl
+++ b/test/nat_SUITE.erl
@@ -108,7 +108,7 @@ server(_Config) ->
         {ok, Since, Port, ExtPort, MockLease}
     end),
     meck:expect(nat, get_external_address, fun(_Context) ->
-        {ok, "127.0.0.1"}
+        {ok, "67.192.56.22"}
     end),
     meck:expect(nat, delete_port_mapping, fun(_Context, tcp, _Port, _Port) ->
         ok
```

Vagabond:
Other than that last broken test, this looks good (squashing to a single commit for clarity would be nice).

Vagabond:
I merged this manually, thanks!

PR #371, opened 2021-08-12
try making this more async

Vagabond:
make the workers accounting use maps

raise accept timeout

don't contend on higher level sup to find worker sup

rework a place I missed

monitor session dials

more hax

add sidejob

rebar stuff

stop inbound workers anyway

shutdown rather than normal

stop worker properly

stop worker out of process, catch closes

inbound workers stop on failure

don't copy state around

remove some debug loggin

fix compilation

Send all new peers to any seed nodes we're connected to

handle sidejob overload

Cherry pick async identify

Fix starting inbound worker

rework peerbook repair and clearing

Clear the peerbook if it's not corrupted

Avoid p2p address conversions in gossip server

Stungun improvements:

Check for 1:1 port mappings with stungun
========================================

Often there's a port mapping manually configured such that the external
port is mapped through to the same internal port. When we detect
symmetric NAT, additionally try to use stungun on these fixed ports to
try to discover if this kind of mapping exists.

Improve stungun retries
=======================

Stungun will now try to continue resolving the NAT state until it gets
an answer (nodes will no longer get stuck in 'unknown').

Improved validation
===================

Stungun now checks the session validating the external IP/Port comes
from an inbound session and ignores it otherwise.

Don't use match_delete on remove_pid, soft delete and use a GC

Do some more work upfront to speed up gc

We take a hit on potentially trying to delete some keys that don't
exist. Deleting against explicit pids means we won't delete anything
not yet marked. :fingers_crossed:

Re-work ETS GC to match all the keys it needs to and be single-pass

Handle overload when starting outbound gossip workers

Yolo spawn

Raise listen backlog

Move the gossip handling out into the worker

Cache gossip handler lookups for 1 minute, infinity timeout

1 minute cache is bad when block times are also 1 minute

Bump splicer to latest release

Fix gossip tuple stuff


src/libp2p_transport_tcp.erl
```diff
@@ -685,7 +685,7 @@ terminate(_Reason, #state{}) ->
 listen_options(IP, TID) ->
     OptionDefaults = [
                       {ip, IP},
-                      {backlog, 1024},
+                      {backlog, 4096},
```
Vagabond, line 688: Make this a config option

PR #379, opened 2021-08-31
Handle race between resolving static nat and port forward

Vagabond:
Even more fixing for #373


Vagabond:
Actually this is an old bug, not related to external IP changes.

vihu (approved):

evanmcc (approved):

PR #386, opened 2021-09-14
Use a configurable external service to verify external IP addresses

Vagabond:
Some times our peers lie to us, so use this as a secondary check to make
sure the discovered address actually makes sense.

related to #284


vihu (approved):
Lgtm

PR #405, opened 2021-11-30
Use a pool of CNAMEs to fetch IPs

jadeallenx:
Problem to solve: As mentioned in https://github.com/helium/miner/pull/1267 we can only have a maximum of 16 IPv4 DNS addresses per CNAME for seed nodes, but we currently have significantly more than 16 seed node IPs in service. 

Solution: Use a pool of DNS CNAMEs to fetch IPs, concatentate them all, and then add the static IPs to form the complete list of seed nodes available for service.  This PR introduces a new environment variable `seed_dns_pool_names` which configures the additional CNAME record names. This variable value is expected to be a list of strings like `["seed2", "seed3", ... ]`

### DIFF
--- a/src/group/libp2p_group_gossip_server.erl
+++ b/src/group/libp2p_group_gossip_server.erl
@@ -44,6 +44,7 @@
 -define(DNS_TIMEOUT, 2000). % millis
 -define(DNS_SLEEP, 100). % millis
 -define(DEFAULT_SEED_DNS_PORTS, [2154, 443]).
+-define(SEED_CONFIG_DNS_NAME, "_seed_config.helium.io").
 
 %% API
 %%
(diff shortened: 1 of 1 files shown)


src/group/libp2p_group_gossip_server.erl
```diff
@@ -44,6 +44,7 @@
 -define(DNS_TIMEOUT, 2000). % millis
 -define(DNS_SLEEP, 100). % millis
 -define(DEFAULT_SEED_DNS_PORTS, [2154, 443]).
+-define(SEED_CONFIG_DNS_NAME, "_seed_config.helium.io").
```
Vagabond, line 47: the DNS record is configutable, can we make this configurable too?

evanmcc: we don't seem to use this anymore now that we pull it from the app env.  should it be the default there, or is there some other place that it's used?

evanmcc (approved):
code looks good, but I had one question about the define

src/group/libp2p_group_gossip_server.erl
```diff
@@ -395,31 +395,62 @@ choose_random_element([E]) -> E;
...
+    case attempt_dns_lookup(CName, ?DNS_RETRIES) of
+        {error, _} = Error ->
+            lager:error("DNS lookup of ~p resulted in ~p after ~p retries", [CName, Error, ?DNS_RETRIES]),
+            Acc;
+        {ok, DNSRecord} ->
+            lager:debug("successful DNS lookup result: ~p for ~p", [lager:pr(DNSRecord, inet), CName]),
+            convert_dns_records(DNSRecord) ++ Acc
     end.
 
+maybe_make_seed_pool_names(Base, []) -> [Base];
```
Vagabond, line 444: This line should look for `undefined` no?

Vagabond: oh, never mind.

src/group/libp2p_group_gossip_server.erl
```diff
@@ -395,31 +395,62 @@ choose_random_element([E]) -> E;
...
+collect_dns_records(Base) ->
+    AdditionalNames = generate_seed_pool_names(application:get_env(libp2p, seed_config_dns_name, undefined)),
+    DNSNames = maybe_make_seed_pool_names(Base, AdditionalNames),
+    lists:foldl(fun do_dns_lookups/2, [], DNSNames).
+
+generate_seed_pool_names(undefined) -> [];
+generate_seed_pool_names(ConfigDnsName) ->
+    case inet_res:lookup(ConfigDnsName, in, txt) of
+        [] -> []; %% there was an error of some kind, return empty list
+        [PoolSizeStr] ->
```
Vagabond, line 420:
```suggestion
        [[PoolSizeStr]] ->
```

```
(blockchain_node@127.0.0.1)1> inet_res:lookup("_seed_config.helium.io", in, txt).
[["4"]]
```

PR #411, opened 2021-12-09
Attempt to throttle arp better, fix peer update gossip

Vagabond:
Allow more ARP requests between seed nodes than any individual user can
do, to allow for secondary ARP to succeed while still limiting user
requests to a reasonable limit.


src/peerbook/libp2p_peer_resolution.erl
```diff
@@ -76,22 +76,32 @@ handle_gossip_data(StreamPid, {_Path, Data}, Handle) ->
         #libp2p_peer_resolution_msg_pb{msg = {response, #libp2p_signed_peer_pb{} = Peer}, re_request=ReRequest} ->
             lager:debug("ARP result for ~p", [libp2p_crypto:pubkey_bin_to_p2p(libp2p_peer:pubkey_bin(Peer))]),
             %% send this peer to the peerbook
+            Res = libp2p_peerbook:put(Handle, [Peer]),
```
jadeallenx, line 79: Is this some kind of caching layer?

jadeallenx (approved):
lgtm

Issue #415, opened 2021-12-15
stungun issues: detected NAT type symmetric


ci-work:
running miner as validator, and etls on dedicated servers, using master libp2p, should be fixed single dedicated ip address, but getting stungun `detected NAT type symmetric` and also a random reports of ip address as `172.` ipv4 network addresses, also incorrect.

several instances of this, but example:
```
2021-12-15 21:38:39.803 [info] <0.1280.0>@libp2p_transport_tcp:handle_info:500 retrying stungun with peer "/p2p/112iynLkeTKvPQnYCYhnrZp4sURMS77EHfCfsVUScG9PBmpf92U9"
2021-12-15 21:38:39.811 [info] <0.1280.0>@libp2p_transport_tcp:handle_info:527 stungun detected NAT type symmetric
2021-12-15 21:38:39.811 [info] <0.1356.0>@libp2p_relay_server:handle_cast:152 requested to init relay but we already have one @ <0.8525.1>
```

ci-work:
possibly related, also seeing:
```
2021-12-15 23:01:17.866 [error] exit Ranch listener ["/ip4/95.217.x.x/tcp/2154","/ip4/172.17.0.1/tcp/2154"] terminated with reason: {normal,{sys,terminate,[<0.12582.0>,normal,infinity]}}
```

ci-work:
additional info:

```
2021-12-16 01:26:31.064 [info] <0.1479.0>@libp2p_transport_tcp:handle_info:500 retrying stungun with peer "/p2p/116Sa97proLW3oTBwUKtR1k7yxABCCwE3dnxvDwqhkh7snbEu3z"
2021-12-16 01:26:31.242 [info] <0.1479.0>@libp2p_transport_tcp:handle_info:527 stungun detected NAT type symmetric
2021-12-16 01:26:31.247 [info] <0.1532.0>@libp2p_relay_server:init_relay:242 initiating relay with peer "/p2p/11x5kNN7RKFEfaqe7TRxXc5bGV2Mz8U4LLwXaHK7hNYA4UZTsXe"
```

peer book entry:

```
# ./bin/miner peer book /p2p/116Sa97proLW3oTBwUKtR1k7yxABCCwE3dnxvDwqhkh7snbEu3z
+--------------------------------------------------------+-------------------+------------+-----------+---------+------------+
|                        address                         |       name        |listen_addrs|connections|   nat   |last_updated|
[...]

PR #432, opened 2022-04-22
Try to prevent stungunning relayed peers

mawdegroot:
Filter the list of connected peers to exclude peers that do not have a public address per suggestion of @ci-work in #415 

I don't think relayed nodes are the problem except the subset of relayed nodes that is running behind some CGNAT or similar system. The easiest way to avoid the issue is to stungun just those peers of which it is known they have a public ip address.

I have been running this change for almost 24h now and it hasn't yet picked up a relay address. I will deploy this to one more node and have it run over the weekend to limit the amount of luck involved.

Some potential problems I considered:
1. **What if there are only relayed peers connected to this node?**
    It shouldn't matter since it will start a relay if it fails a few times in a row.

Addresses #415

### DIFF
--- a/src/libp2p_transport_tcp.erl
+++ b/src/libp2p_transport_tcp.erl
@@ -492,7 +492,7 @@ handle_info(stungun_retry, State=#state{observed_addrs=Addrs, tid=TID, stun_txns
             {PeerPath, TxnID} = libp2p_stream_stungun:mk_stun_txn(),
             %% choose a random connected peer to do stungun with
             {ok, MyPeer} = libp2p_peerbook:get(libp2p_swarm:peerbook(TID), libp2p_swarm:pubkey_bin(TID)),
-            case [libp2p_crypto:pubkey_bin_to_p2p(P) || P <- libp2p_peer:connected_peers(MyPeer)] of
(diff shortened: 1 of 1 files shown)


mawdegroot:
I don't think it will matter since if it has a public ip address that will be the prioritized address that you connect to. That it also has a p2p address does not indicate that the peer itself has a broken internet configuration, just that it was hit by the same problem this commit tries to prevent.

I will run a second node with your suggestion and see if it makes any difference regardless 👍

mawdegroot:
The instance with this check has been running over the weekend without starting up a relay so this fix seems to prevent (at least a large part) of the issue. The second instance that also prevents selecting connected peers with a p2p address (as secondary address) achieved the same results so I propose to keep the change minimal and only check for a public IP.

Removing the draft label as the fix has had the intended result over a longer run.

jadeallenx (approved):

Vagabond:
This is actually wrong as `libp2p_peer:connected_peers()` doesn't return peerbook entries itself, but rather pubkey bins, which are not valid to call `libp2p_peer:listen_addrs()` on.

PR #437, opened 2022-06-07
restrict iterator_move to returning valid pubkeys

joecaswell:
Add a helper to filter iterator_move so only 33-byte keys are considered.

### DIFF
--- a/src/peerbook/libp2p_peerbook.erl
+++ b/src/peerbook/libp2p_peerbook.erl
@@ -200,12 +200,43 @@ random(Peerbook, Exclude) ->
 random(Peerbook, Exclude, Pred) ->
     random(Peerbook, Exclude, Pred, 15).
 
+
+iterator_move_filtered(KeyFun, Iterator, Action) ->
+    case rocksdb:iterator_move(Iterator, Action) of
+        {ok, Key, Value} ->
+            case catch KeyFun(Key) of
+                true -> {ok, Key, Value};
+                _ ->
+                    NextAction = case Action of
+                                     first -> next;
+                                     next -> next;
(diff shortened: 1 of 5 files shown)


Vagabond:
Might be good to add a quick test for this?

src/peerbook/libp2p_peerbook.erl
```diff
@@ -200,12 +200,43 @@ random(Peerbook, Exclude) ->
 random(Peerbook, Exclude, Pred) ->
     random(Peerbook, Exclude, Pred, 15).
 
+
+iterator_move_filtered(KeyFun, Iterator, Action) ->
+    case rocksdb:iterator_move(Iterator, Action) of
```
anthonyra, line 205: Also fwiw, if the iterator is built with seek option. This could loop back to the beginning and when it does that the guard pattern changes. It flips the key and value. Atleast from what I’ve experienced.

Vagabond: that seems bizarre?

evanmcc (approved):

src/peerbook/libp2p_peerbook.erl
anthonyra: I tried to look to see why this happens in the source for rocksDB but it doesn't jump out to me. I just encountered this while working on my ETL optimizations that I had done not to long ago.

joecaswell: The docs have some caveats about undefined behavior around using seek iterator when the database or column family was opened with a prefix option, I don't think we are using the prefix option for peerbook though.

PR #439, opened 2022-06-09
Remove old code that upgraded old style peerbooks

Vagabond:
This code has been unnecessary for a very long time and was causing
memory issues when invalid peerbook keys were encountered on valid
peerbooks.


ke6jjj:
LGTM!

jeffgrunewald (approved):

PR #440, opened 2022-06-13
network_id is wrapped in its own list

jadeallenx:
@jontow found that the network id assigned in miner was not being correctly passed through to libp2p because the network id itself is wrapped in a proplist that needs to be exacted first.


jontow:
:+1:

jontow (approved):

PR #441, opened 2022-06-15
always match peer to self netid

jeffgrunewald:
assuming we've defined network ids on nodes where it really matters, this change would continue to ensure that they only peer up with nodes on the same network. however, if the network id is `undefined` we aren't picky about peering up with others where the network id is also `undefined`


madninja (approved):

jadeallenx (approved):

jadeallenx:
OK, closing for #442

PR #444, opened 2022-07-07
Jc/local peer get exception

joecaswell:
Return the local peer record even if the network is undefined.
If it doesn't exist or has a different network id than the swarm, recreate the local peer record first


Vagabond:
@joecaswell can you fix the merge conflict?

joecaswell:
rebased on master

PR #447, opened 2022-07-19
Various dialyzer fixes

Vagabond:


src/libp2p_config.erl
```diff
@@ -236,7 +236,7 @@ insert_listener(TID, Addrs, Pid) ->
 lookup_listener(TID, Addr) ->
     lookup_pid(TID, ?LISTENER, Addr).
 
--spec remove_listener(ets:tab(), string()) -> true.
+-spec remove_listener(ets:tab(), string() | undefined) -> true.
```
Vagabond, line 239: ideally we remove the calls when the listener addr is undefined instead

src/libp2p_config.erl
```diff
@@ -423,7 +423,8 @@ lookup_group(TID, GroupID) ->
 remove_group(TID, GroupID) ->
     remove_pid(TID, ?GROUP, GroupID).
 
--spec all_groups(ets:tab()) -> [{string(), pid()}].
```
Vagabond, line 426: This is not the same spec

src/peerbook/libp2p_peerbook.erl
```diff
@@ -428,8 +428,9 @@ lookup_association(Handle=#peerbook{}, AssocType, AssocAddress) ->
 %% Gossip Group
 %%
 
--spec handle_gossip_data(pid(), inbound | seed | peerbook, string(), {string(), [libp2p_peer:peer()]}, peerbook()) -> noreply.
-handle_gossip_data(_StreamPid, _Kind, _Peer, {"gossip/1.0."++_, DecodedList}, Handle) ->
+-spec handle_gossip_data(pid(), inbound | seed | peerbook, string(), {string(), binary()}, peerbook()) -> noreply.
+handle_gossip_data(_StreamPid, _Kind, _Peer, {"gossip/1.0."++_, Bin}, Handle) ->
```
Vagabond, line 432: we should figure out if the first element of the tuple is a string or a binary

src/libp2p_config.erl
joecaswell: Since the state record permits address to be undefined, that could be anywhere it is passed unchecked, like https://github.com/helium/erlang-libp2p/blob/master/src/relay/libp2p_relay_server.erl#L199

